PR #106, opened 2015-03-24
Add docker support

emilevauge:
Hi

I thought it would be nice to have a Docker image of Howdoi. This allows users who don't have python installed to use this utility anyway.

Emile

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,10 @@
+FROM python:3-onbuild
+MAINTAINER emile@vauge.com
+
+ADD . /howdoi
+WORKDIR /howdoi
+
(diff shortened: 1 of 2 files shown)


gleitz:
How does the docker command find out about this package? Does it need to be registered somewhere?

EmileVauge:
In my example `docker pull emilevauge/howdoi`, it gets the Docker image from this public repository: https://registry.hub.docker.com/u/emilevauge/howdoi/
If you create an account on Docker hub and register your GitHub project gleitz/howdoi, Docker Hub will automatically build a new Docker image for each commit on your Github repo (based on the Dockerfile provided in my PR).

Issue #111, opened 2015-03-30
Exception for certain queries using Python 2.9


giacomocusinato:
I get an exception for certain queries, like the one that follows:

![hoedoi_exception](https://cloud.githubusercontent.com/assets/7659518/6907749/1ef3b42e-d739-11e4-97d2-59fb731a351e.png)

I get this error just using Python 2.9, with Python 3.4 works just fine.
This block of code is the one throwing the exception:

```python
try:
        return requests.get(url, headers={'User-Agent': random.choice(USER_AGENTS)}, proxies=get_proxies()).text
    except requests.exceptions.SSLError as e:
        print('[ERROR] Encountered an SSL Error. Try using HTTP instead of '
              'HTTPS by setting the environment variable "HOWDOI_DISABLE_SSL".\n')
        raise e
`
``

Using `urllib2` module instead `requests` could solve the problem. I'll try to submit a pull request.

gleitz:
I would rather keep the requests library. If there are issues with it please report them to https://github.com/kennethreitz/requests

PR #112, opened 2015-03-30
Replaced requests module with urllib2 for issue #111

giacomocusinato:
Works now fine for me with both Python 2.9 and 3.4

### DIFF
--- a/howdoi/howdoi.py
+++ b/howdoi/howdoi.py
@@ -13,10 +13,10 @@ import glob
 import os
 import random
 import re
-import requests
 import requests_cache
 import sys
-from . import __version__
+import urllib2
+# from . import __version__
 
 try:
     from urllib.parse import quote as url_quote
@@ -34,8 +34,6 @@ from pygments.formatters import TerminalFormatter
 from pygments.util import ClassNotFound
 
 from pyquery import PyQuery as pq
-from requests.exceptions import ConnectionError
-from requests.exceptions import SSLError
 
 # Handle unicode between Python 2 and 3
 # http://stackoverflow.com/a/6633040/305414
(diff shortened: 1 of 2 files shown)


gleitz:
Thanks for the pull request! Please remove requests from `requirements.txt` if is isn't needed anymore.

giacomocusinato:
This should be it :)

howdoi/howdoi.py
```diff
@@ -13,7 +13,6 @@
 import os
 import random
 import re
-import requests
 import requests_cache
```
gleitz: We use requests_cache to speed up query times. Is there a similar library for urllib2?

howdoi/howdoi.py
```diff
@@ -83,9 +85,14 @@ def get_proxies():
...
 def get_result(url):
     try:
-        return requests.get(url, headers={'User-Agent': random.choice(USER_AGENTS)}, proxies=get_proxies()).text
-    except requests.exceptions.SSLError as e:
-        print('[ERROR] Encountered an SSL Error. Try using HTTP instead of '
+        urllib2.install_opener(
+            urllib2.build_opener(
+                urllib2.ProxyHandler(get_proxies())
+            )
+        )
```
gleitz: I think the opener only needs to be installed once, correct? Perhaps that can happen at the top of the file.

requirements.txt
```diff
@@ -3,5 +3,4 @@ argparse==1.2.1
 cssselect==0.7.1
 lxml==3.2.4
 pyquery==1.2.6
-requests==2.0.1
 requests-cache==0.4.4
```
gleitz: Let's figure out a replacement for requests-cache

howdoi/howdoi.py
griesmey: What would be the advantage of using the urllib2 package here?  The requests packages help with readability and is cleaner.

gleitz:
I would rather keep the requests library.

PR #132, opened 2016-02-18
PEP 8 conventions

tusharmakkar08:
- Private functions named properly
- `sys.version` is used explicitly instead of  `ImportError` (`Explicit is better than implicit`) :grinning: 
- Line length made <= 120 char

### DIFF
--- a/howdoi/howdoi.py
+++ b/howdoi/howdoi.py
@@ -78,7 +78,7 @@ def get_proxies():
     return filtered_proxies
 
 
-def get_result(url):
+def _get_result(url):
     try:
         return requests.get(url, headers={'User-Agent': random.choice(USER_AGENTS)}, proxies=get_proxies()).text
     except requests.exceptions.SSLError as e:
@@ -87,19 +87,19 @@ def get_result(url):
         raise e
 
 
-def is_question(link):
+def _is_question(link):
     return re.search('questions/\d+/', link)
 
 
-def get_links(query):
-    result = get_result(SEARCH_URL.format(URL, url_quote(query)))
+def _get_links(query):
+    result = _get_result(SEARCH_URL.format(URL, url_quote(query)))
     html = pq(result)
     return [a.attrib['href'] for a in html('.l')] or \
         [a.attrib['href'] for a in html('.r')('a')]
 
 
 def get_link_at_pos(links, position):
-    links = [link for link in links if is_question(link)]
+    links = [link for link in links if _is_question(link)]
     if not links:
         return False
 
@@ -110,7 +110,7 @@ def get_link_at_pos(links, position):
     return link
 
 
-def format_output(code, args):
+def _format_output(code, args):
     if not args['color']:
         return code
     lexer = None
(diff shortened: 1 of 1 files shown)


gleitz:
Thanks @tusharmakkar08! I'll review and come back with thoughts

howdoi/howdoi.py
```diff
@@ -37,14 +27,18 @@
 from requests.exceptions import ConnectionError
 from requests.exceptions import SSLError
 
-# Handle unicode between Python 2 and 3
-# http://stackoverflow.com/a/6633040/305414
```
gleitz: Let's keep this stackoverflow link – it helps explain why we're defining the `u` function.

howdoi/howdoi.py
```diff
@@ -18,16 +18,6 @@
...
-try:
-    from urllib.parse import quote as url_quote
-except ImportError:
-    from urllib import quote as url_quote
-
-try:
-    from urllib import getproxies
-except ImportError:
-    from urllib.request import getproxies
-
```
gleitz: Nice – I like what you did below. Much more explicit.

howdoi/howdoi.py
```diff
@@ -59,8 +53,10 @@ def u(x):
 USER_AGENTS = ('Mozilla/5.0 (Macintosh; Intel Mac OS X 10.7; rv:11.0) Gecko/20100101 Firefox/11.0',
                'Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:22.0) Gecko/20100 101 Firefox/22.0',
                'Mozilla/5.0 (Windows NT 6.1; rv:11.0) Gecko/20100101 Firefox/11.0',
-               'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_7_4) AppleWebKit/536.5 (KHTML, like Gecko) Chrome/19.0.1084.46 Safari/536.5',
-               'Mozilla/5.0 (Windows; Windows NT 6.1) AppleWebKit/536.5 (KHTML, like Gecko) Chrome/19.0.1084.46 Safari/536.5',)
+               'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_7_4) AppleWebKit/536.5 (KHTML, like Gecko) '
+               'Chrome/19.0.1084.46 Safari/536.5',
```
gleitz: It's a little hard to read the multiline string concatenation, especially if you miss the comma. How about [python line continuation](http://stackoverflow.com/a/1874679) using parenthesis instead?

tusharmakkar08: Parenthesis seems to be a nice solution. :+1:

howdoi/howdoi.py
tusharmakkar08: Ok. I thought it was implicit while mentioning the `u` function. I will update the pull request correspondingly.

tusharmakkar08:
I have updated the pull request with the necessary changes.

howdoi/howdoi.py
tusharmakkar08: :+1:

gleitz:
Put the stackoverflow link back in and I'll merge

gleitz:
Oh nvm I see it. Thanks!

Issue #134, opened 2016-02-22
HOWDOI_COLORIZE doesn't automatically set color


jackmaney:
I'm using OS X 10.9.5 with `howdoi` version 1.1.7 on Python 3.5.1.

I've enabled the `HOWDOI_COLORIZE` variable in my `.bash_profile`:

```
$ tail -n 1 ~/.bash_profile
export HOWDOI_COLORIZE=1
```

However, setting this doesn't seem to have any effect:

![screen shot 2016-02-22 at 12 22 07 pm](https://cloud.githubusercontent.com/assets/1554625/13228061/5c970c94-d95f-11e5-8421-427eba0c6bd9.png)

Is there something obvious that I'm missing, here?

Thank you for your time.

gleitz:
Looks like the output is colorized, no? The DATE and backticks have color

gleitz:
Ah, I see. The environment variable isn't being used. Feel free to submit a pull request.

PR #135, opened 2016-02-23
Fixed broken import and made .gitignore more robust

jackmaney:
The main culprit seemed to be the import of `TerminalFormatter`, as it's in `pygments.formatters.terminal`, not `pgyments.formatters`. So, after changing that import, the issues in #134 behind automatic colorization via the `HOWDOI_COLORIZE` environment variable seem to have been fixed.

I also added some stuff to the `.gitignore` file (mostly via [gitignore.io](https://www.gitignore.io/)) to make the ignoring a little more robust. I needed to add at least *something*, since I was using PyCharm.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,187 @@
-*.pyc
-.DS_Store
-dist/
+
+# Created by https://www.gitignore.io/api/python,osx,windows,linux,pycharm
+
+### Python ###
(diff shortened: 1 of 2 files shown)


gleitz:
Thanks! Can we also bump Pygments to 2.1.1?

Also, is it possible to add something to `test_howdoi.py` that catches this error?

PR #138, opened 2016-03-03
Using pythonic ways for some statements

swapnilsm:


gleitz:
Thanks for the request. Please make sure all tests pass before I'll review.

swapnilsm:
The dict comprehension syntax used was only supported on 2.7+. Now I have fixed it to support 2.6 as well. And the build passes - https://travis-ci.org/swapnilsm/howdoi/builds/113586408

gleitz:
Thanks for the pull request but I think the `dict` is actually a little harder to read. I like the step-by-step logic of the if/else

PR #140, opened 2016-03-08
Added '-u/--url option to set URL

rcmdnk:
Added new option in the parser to set URL.

### DIFF
--- a/howdoi/howdoi.py
+++ b/howdoi/howdoi.py
@@ -225,6 +225,7 @@ def get_parser():
     parser.add_argument('-n', '--num-answers', help='number of answers to return', default=1, type=int)
     parser.add_argument('-C', '--clear-cache', help='clear the cache',
                         action='store_true')
+    parser.add_argument('-u', '--url', help='Set origin URL (default: ' + URL + ')', default=URL, type=str)
     parser.add_argument('-v', '--version', help='displays the current version of howdoi',
                         action='store_true')
     return parser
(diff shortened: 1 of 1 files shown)


gleitz:
Thanks for the pull request but you can already set the URL using the [HOWDOI_URL](https://github.com/rcmdnk/howdoi/blob/arg_url/howdoi/howdoi.py#L52) environment variable.

rcmdnk:
Thank you for the response. 
Though I know that value, I was thinking that it is useful if I can set it as an argument, too.
But I respect your decision.

Anyway, thank you for the useful tool!

gleitz:
Yes I try to keep the command line clean. If you want, you can do something like this:

`function howdoi-arqade { export HOWDOI_URL="gaming.stackexchange.com" && howdoi "$@"; }`

```
$ howdoi-arqade who is ness in super smash bros
As you know, Ness made his debut in the game EarthBound for the SNES. Now one of the PSI abilities that he, as well as another player named Poo, can learn is called PSI Teleport. As the name suggests, this PSI power gives you the ability to warp to places you've already visited. There are two versions of this power-up, PSI Teleport α (you warp by running in a straight line), and PSI Teleport β (you warp by running in a circle). Now the problem with PSI Teleport is that if you happen to run into something while performing the move, you won't teleport. When you hit something, Ness and his friends appear to be burnt. THIS is what you see when Ness makes his entrance onto the stage in Smash Bros. Brawl: it's the same effect that occurs if you run into something while attempting PSI Teleport.
I hope this helps :)
```

PR #141, opened 2016-03-08
Added show-url option

rcmdnk:
New option "-s"/"--show-url" to show the source URL of the answer.


rcmdnk:
There is "all" option and it seems enough.
sorry for the spam.

gleitz:
You can show the URL with `-l`

PR #143, opened 2016-03-15
_is_question logic check is moved outside

abbi031892:
I was going through the code and found that is question was getting checked in the method giving links at position, so i moved it outside, do you think the same or there is something i am missing ? I will change the tests later, but first want to confirm with my thought.


gleitz:
Sure, you can filter out questions in `_get_answer` before passing to `get_link_at_pos`. It should also start with an `_` so please include that in the change.

abbi031892:
Yes now i did as you asked, and also travis build is passing, please check it.

Issue #152, opened 2016-08-15
Include the answer link in the output by default


jericson:
By default, answers provided by `howdoi` come from Stack Overflow, which has an [attribution policy](http://blog.stackoverflow.com/2009/06/attribution-required/). It's pretty typical to include a comment with a link to the SO answer when using code. (Even this project [follows that convention](https://github.com/gleitz/howdoi/blob/master/howdoi/howdoi.py#L36).) If you use the `-l` option, you get a link to the answer (and no code), but there's no way to get the link _and_ the code. 

My suggestion is to keep the `-l` option the same and add a URL to the default output. This would also make the multiple answer output more useful:

```
$ howdoi copy table lua -n 2
--- http://stackoverflow.com/questions/37565836/copy-table-elements-to-new-table ---
table.insert(MainTable, TempTable)

--- http://stackoverflow.com/questions/4535152/cloning-a-lua-table-in-lua-c-api ---
shallow_copy = function(tab)
    local retval = {}
    for k, v in pairs(tab) do
        retval[k] = v
    end
    return retval
end
```

I think that looks a lot better than "Answer 1" and "Answer 2".

(Disclosure: I work as a Community Manager [`howdoi community manager`] for Stack Overflow.)

gleitz:
Very nice suggestion, @jericson. I would be happy to accept a pull request with the URL included.

gleitz:
Fixed in #169

PR #153, opened 2016-08-16
Include urls by default.

jericson:
Addresses https://github.com/gleitz/howdoi/issues/152


gleitz:
Thanks @jericson.

I wonder if we should use `//` to preface the URL instead of surrounding it with `---` as it will be more likely to already be in comment form.

jericson:
That seems reasonable too. I was also thinking of `#`, but the important thing is to separate it from the answer somehow. (Text is tricky!)

GabLeRoux:
What about some special character?

    echo '\u2605'
    ★

Since results may already have comments (which differs from languages to languages).

tra38:
Any move on merging in this PR into master?

gleitz:
Closed in favor of #169

PR #166, opened 2017-09-06
add command line shortcut

eoglethorpe:


gleitz:
Looks cool! Would you check on the tests? They may need to be updated.

README.rst
```diff
@@ -97,8 +97,20 @@ Usage
                             number of answers to return
       -C, --clear-cache     clear the cache
       -v, --version         displays the current version of howdoi
+      
+      
+As a shortcut, if you commonly use the same paremeters each time and don't want to type them, add something similar to your .bash_profile (or otherwise). This example gives you 5 colored results each time.
 
+::
+
+    alias h='function hdi(){ howdoi $* -c -n 5; }; hdi'
```
gleitz: I generally use the following which is a little easier to read:
`alias h='howdoi -c -n 5'`
Any benefit to your functional version?

eoglethorpe: yo - not sure about your alias to be honest and my grasp of how they work is a bit limited. I can see that it works but I don't understand how to command is passed in?

the upswing i can think of my method is that you can start to add in parameter inputs for function inputs like number of responses to give if you were to add that but otherwise seems the same as yours

eoglethorpe: sure thing, they are fixed with #167.

thanks for the library btw, has been a good addition to my work. lmk if you have anything else i can help out with

PR #169, opened 2017-09-17
Add links, with ★ characters to separate the link from the content

tra38:
This is my attempt to resolve Issue #152, using PR #153 as a starting point. I fixed the merge conflict that prevented merger while also adding in the ★ character as suggested by [GabeLeRoux]( https://github.com/gleitz/howdoi/pull/153#issuecomment-244577201). Let me know if you like it!

### DIFF
--- a/howdoi/howdoi.py
+++ b/howdoi/howdoi.py
@@ -60,7 +60,8 @@ USER_AGENTS = ('Mozilla/5.0 (Macintosh; Intel Mac OS X 10.7; rv:11.0) Gecko/2010
                 'Chrome/19.0.1084.46 Safari/536.5'),
                ('Mozilla/5.0 (Windows; Windows NT 6.1) AppleWebKit/536.5 (KHTML, like Gecko) Chrome/19.0.1084.46'
                 'Safari/536.5'), )
-ANSWER_HEADER = u('--- Answer {0} ---\n{1}')
+STAR_HEADER = u('\u2605')
+ANSWER_HEADER = u('{2}  {0} {2}\n{1}')
 NO_ANSWER_MSG = '< no answer given >'
 XDG_CACHE_DIR = os.environ.get('XDG_CACHE_HOME',
                                os.path.join(os.path.expanduser('~'), '.cache'))
@@ -158,6 +159,7 @@ def _get_answer(args, links):
     args['tags'] = [t.text for t in html('.post-tag')]
 
     if not instructions and not args['all']:
+        texts = []
         text = first_answer.find('.post-text').eq(0).text()
     elif args['all']:
         texts = []
@@ -168,7 +170,7 @@ def _get_answer(args, links):
                     texts.append(_format_output(current_text, args))
                 else:
                     texts.append(current_text)
-        texts.append('\n---\nAnswer from {0}'.format(link))
+        texts.append('\n---\n'.format(link))
         text = '\n'.join(texts)
     else:
         text = _format_output(instructions.eq(0).text(), args)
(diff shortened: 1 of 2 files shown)


gleitz:
This is cool. I often pipe howdoi directly to the clipboard to paste into code, and I don't like that I'll now have to remove the line with the ★.
What do you think about only displaying the link if you ask for more than one answer?
I suppose the other option is to have the links + stars only visible when a flag is set, but we already have so many flags it might be overkill.

tra38:
Done. I also allowed links to be displayed if the user uses the "all" flag, simply because the original behavior of "all" is to display the link along with the full answer.

howdoi/howdoi.py
```diff
@@ -158,6 +159,7 @@ def _get_answer(args, links):
     args['tags'] = [t.text for t in html('.post-tag')]
 
     if not instructions and not args['all']:
+        texts = []
```
gleitz: Don't think this line is necessary

tra38: Good catch. Removed.

gleitz:
Hey, I noticed that this code change breaks python3. I wonder if you know how to fix, by allowing Python3 to print ★?

```
$ python3 -m howdoi.howdoi format date bash -n2
Traceback (most recent call last):
  File "/usr/local/Cellar/python3/3.6.2/Frameworks/Python.framework/Versions/3.6/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/local/Cellar/python3/3.6.2/Frameworks/Python.framework/Versions/3.6/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/Users/bgleitzman/projects/howdoi/howdoi/howdoi.py", line 279, in <module>
    command_line_runner()
  File "/Users/bgleitzman/projects/howdoi/howdoi/howdoi.py", line 275, in command_line_runner
    print(howdoi(args))
UnicodeEncodeError: 'ascii' codec can't encode character '\u2605' in position 0: ordinal not in range(128)
```

tra38:
I'm currently trying to replicate the issue myself. Right now, Python3 seems to handle that query just fine on my computer...

tra38:
Found this, while trying to replicate the issue.

```
★  Answer from https://stackoverflow.com/questions/20923663/unicodeencodeerror-ascii-codec-cant-encode-character-in-position-0-ordinal ★
When Python prints and output, it automatically encodes it to the target medium. If it is a file, UTF-8 
will be used as default and everyone will be happy, but if it is a terminal, Python will figure out the 
encoding the terminal is using and will try to encode the output using that one.
This means that if your terminal is using ascii as encoding, Python is trying to encode scissor char to 
ascii. Of course, ascii doesn't support it so you get Unicode decode error.
This is why you always have to explicitly encode your output . Explicit is better than implicit 
remember? To fix your code you may do:
import sys
sys.stdout.buffer.write(chr(9986).encode('utf8'))
This seems a bit hackerish. You can also set PYTHONIOENCODING=utf-8 before executing the 
script. I'am uncomfortable with both solutions. Probably your console doesn't support utf-8 and you 
see gibberish. But your program will be behaving correctly.
What I strongly recommend if you definitely need to show correct output on your console is to set 
your console to use another encoding, one that support scissor character. (utf-8 perhaps). On 
Linux, that can be achieve by doing: export lang=UTF_8 . On Windows you change the console's 
code page with chcp . Just figure out how to set utf8 in yours and IMHO that'll be the best solution.
print
sys.stdout.write
sys.stdout.buffer.write(("|\t "+ chr(9986) +" PySnipt'd " + chr(9986)+" \t|").encode('utf8'))
I suggest you to take a read at the docs to see what's going on under the hood with print function 
and with sys.stdout : http://docs.python.org/3/library/sys.html#sys.stdin
Hope this helps!
```

This SO answer could probably explain the issue, but for me to verify if this is correct, I'll still need to set my own terminal's encoding to use ASCII (right now, it appears to default to UTF-8).

gleitz:
Thanks for the update. I believe this will fail every time:
`PYTHONIOENCODING=ascii python3 -m howdoi.howdoi format date bash -n3`

And this will succeed:
`PYTHONIOENCODING=utf-8 python3 -m howdoi.howdoi format date bash -n3`

I think you're right that a fix is replacing the `print` statement with that `sys` command from above. See https://github.com/gleitz/howdoi/pull/178.

tra38:
Cool! Glad that got fixed. :+1:

PR #171, opened 2017-10-15
Added another installation method

mehandes:
Added installation method utilizing Nix package manager


gleitz:
I think the existing 4 installation methods are enough, but thanks for the suggestion.

PR #174, opened 2017-11-07
update formula to work with linuxbrew

emanresusername:
i assume it'd also work with regular homebrew... but couldn't test that :thinking: 
would you want to keep maintaining it here, or turn it over to the homebrew repo? 
(or is it already maintained in a different tap somewhere that slipped under my radar?)

### DIFF
--- a/howdoi.rb
+++ b/howdoi.rb
@@ -1,23 +1,56 @@
 require 'formula'
 
 class Howdoi < Formula
+  include Language::Python::Virtualenv
+
   homepage 'https://github.com/gleitz/howdoi/'
   url 'https://pypi.python.org/packages/23/2d/9790707eed08c802daee32183e7c98ec2e9797564dad229738b7f178e18e/howdoi-1.1.9.tar.gz'
   sha256 '2b4a0ebc39c9e0489a17ef73e3eec7dd4a2975e07b995406887c639481da2bb5'
 
+  depends_on :python if MacOS.version <= :snow_leopard
+  depends_on "libxslt" unless OS.mac?
+
+  resource "Pygments" do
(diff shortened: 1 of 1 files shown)


gleitz:
The last time I looked into this, Homebrew did not want to host things that were also available via `pip` (see [here](https://github.com/Homebrew/brew/blob/master/docs/Acceptable-Formulae.md)), so I decided to host my own formula.
Thanks for updating this. I would like to test this formula with regular homebrew just to make sure it works.

emanresusername:
oh interesting, i didn't know the history 
well if the formulas i was looking at when i remembered this are any indication ([magic-wormhole](https://github.com/Linuxbrew/homebrew-core/blob/master/Formula/magic-wormhole.rb#L2) and [mitmproxy](https://github.com/Linuxbrew/homebrew-core/blob/master/Formula/mitmproxy.rb#L199)) 
looks like that policy might've changed

idk if the `virtualenv` stuff is a relatively recent development, 
i could imagine some cases against using homebrew to install a bunch of stuff with to the same global `pip` without the `virtualenv`s :thinking: 

in any case, yeah, lmk if it works with regular homebrew

gleitz:
Just installed this on a mac and it worked, so I think we're all good to go. Thx!

emanresusername:
dope 👌 
i'll raise an issue with homebrew-core and reference this and see if they want to take the maintenance under their wing ...

emanresusername:
@gleitz it's in homebrew-core officially now https://github.com/Homebrew/homebrew-core/commit/5aefbfe9fe2b488cb39815dc70ea96a947a20006

Issue #175, opened 2017-11-23
Wrong links get displayed when typing "howdoi cat --all -n 3"


tra38:
```
>howdoi cat --all -n 3
★  Answer from http://rads.stackoverflow.com/amzn/click/B000RIA95G ★
While cat does stand for "concatenate", what it actually does is simply display one or multiple files, in order of their appearance in the command line arguments to cat . The common pattern to view the contents of a file on Linux or *nix systems is:
cat <file>
The main difference between cat and Git's cat-file is that it only displays a single file (hence the -file part). Git's cat-file doesn't really stand for "concatenate"; it simply is a reference to the behavior of the cat command.
git-cat-file - Provide content or type and size information for repository objects
Technically, you can use git cat-file to concatenate files, if you use Batch Output mode:
BATCH OUTPUT If --batch or --batch-check is given, cat-file will read objects from stdin, one per line, and print information about them. By default, the whole line is considered as an object, as if it were fed to git-rev-parse [1].

★  Answer from http://rads.stackoverflow.com/amzn/click/0679449531 ★
the problem is that you are using a Fileconnection in combination with cat then the append won't work. There are several option you could use, the most easy one is to this: first "create" the file, if you want to add a header for example:
header = "some header" 
## if you don't want to use a header then leave the header blank
header =""
cat(text, file="outputR.txt", append=FALSE, sep = "\n")
notice the append = FALSE this is necessary if you want to clear the file if it already exist otherwise you have to use append = TRUE the you can write text to it using:
text = text = paste(toString(id), ":", toString(goIDs))
cat(text file="outputR.txt", append=TRUE, sep = "\n")

★  Answer from http://rads.stackoverflow.com/amzn/click/B003E6WSHK ★
$ cat > hello.c
#include <stdio.h>

main(void) {
    printf("Hello, world.\n");
}
^D
$ gcc hello.c
$ ./a.out
Hello, world.
```

If I request the actual links though...
```
>howdoi cat -l -n 3
https://stackoverflow.com/questions/38177026/what-does-cat-file-stand-for-in-git

https://stackoverflow.com/questions/39942297/r-doesnt-append-lines-to-file-using-cat-or-write

https://stackoverflow.com/questions/921455/how-do-i-program-using-cat
```

gleitz:
That's odd, I get the correct links.

```
>howdoi cat --all -n 3
★  Answer from http://rads.stackoverflow.com/amzn/click/B000RIA95G ★
[...]

PR #176, opened 2017-11-23
Fix issue where the wrong answer url was being displayed to the user

tra38:
Issue was originally reported in #175. Since we're dependent on ```_get_questions``` to extract the question URLs from the list of all URLs, I wrote an automated test for that specific function as well.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -121,7 +121,7 @@ Development
 -----------
 
 -  Checkout the repo
--  Run `python -m howdoi.howdoi QUERY` (if you try running `python howdoi/howdoi.py` you my get `ValueError: Attempted relative import in non-package`).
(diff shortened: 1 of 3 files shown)


tra38:
The first commit I made failed on Python 2.6 since I used ```assertSequenceEqual``` in my automated test, a method that was only introduced in Python 2.7. In my haste to try to support Python 2.6, I made two commits that ended up failing on *all* versions. The last commit finally passes TravisCI, but I did wind up creating a rather messy commit history in the process.

Of course, the bigger question I have is whether Python 2.6 should even be supported in the first place. But if we do want to support it, at least our automated tests should pass.

gleitz:
Thanks for the pr. Would you squash all of those commits into one so it can make the version history look a little better?

tra38:
Sure, I'll squash these commits into one. Give me a second to open up a correct PR with the squashed commits.

PR #177, opened 2017-11-23
Fix issue where the wrong answer url was being displayed to the user

tra38:
Also added an automated test to check if _get_questions is working.
Ensured that this automated test works on Python 2.6 as well.

Issue was originally reported in #175, and fixed in #176, but a problem with rebasing forced me to abandon the PR and start a new "clean" PR. This time, everything's good to go. 👍


gleitz:
Awesome!

PR #181, opened 2017-12-14
Add support for bing search engine

WindSoilder:
Sadly found that howdoi doesn't work for me, after hacking into the code, I found that howdoi use *Google* to search for relative links of *StackOverflow*, but *Google* search engine doesn't support for my country :(

So I would like to add support for *Bing* search engine.  To make it work in my environment.

When user input command like this:
    howdoi usage of tar -b
Howdoi will use bing search engine to find stackoverflow links, and then user can get the result here :)

Consider for add an argument rather than use environment variable:
   I think that use argument for bing searching can make user easily select when to use bing or google, if using environment variable, it's a little more complicated.

And by the way, I have add 2 blank line to make our code more fitted with pep8

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -80,7 +80,7 @@ Usage
 
 ::
 
-    usage: howdoi.py [-h] [-p POS] [-a] [-l] [-c] [-n NUM_ANSWERS] [-C] [-v] QUERY [QUERY ...]
+    usage: howdoi.py [-h] [-p POS] [-a] [-l] [-c] [-b] [-n NUM_ANSWERS] [-C] [-v] QUERY [QUERY ...]
 
     instant coding answers via the command line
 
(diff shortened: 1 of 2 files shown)


Omar-Elrefaei:
Good job, I just have a note on the use of environment variables. It would be also nice to have a variable control the search engine as it would be far easier for long-term use. I think they both are different tools for different use cases.

Though one can use an alias to automatically use the argument every time, but still env vars feels the more unix-y way

WindSoilder:
Emm...agree with that.
By the way, I think if using environment variable to decide using *Bing* or *Google*, we can leave the *-b* argument in the argument list.
Because if existed environment variable, using argument can be redundant.

gleitz:
I support configuring the search engine through an environment variable. Perhaps:

`HOWDOI_SEARCH_ENGINE=bing` (with a default to `google`). Also allows for different search engines in the future.

WindSoilder:
Done with it.

Issue #186, opened 2018-02-06
Answers being printed without newlines


BackwardSpy:
When I run `howdoi`, the answer is compressed into a single line.

Example in fish shell:
```shell
~ howdoi -c python json decode
import json j = json.loads('{"one" : "1", "two" : "2", "three" : "3"}') print j['two']
```

Example in bash:
```shell
chris@galactica:~$ howdoi -c python json decode
import json j = json.loads('{"one" : "1", "two" : "2", "three" : "3"}') print j['two']
```

I have tried add/removing various flags to see if that makes a difference, but the behaviour stays consistent:

```shell
~ howdoi python json decode
import json j = json.loads('{"one" : "1", "two" : "2", "three" : "3"}') print j['two']
~ howdoi -a python json decode
★  Answer from https://stackoverflow.com/questions/7771011/parse-json-in-python ★
Very simple:
import json j = json.loads('{"one" : "1", "two" : "2", "three" : "3"}') print j['two']
```

It is not just this specific query that acts this way, every question I enter behaves like this.

It seems that all the `<pre>` blocks in each answer are being compressed into a single line. In the case where an answer contains multiple `<pre>` blocks, they each appear on separate lines.

I think this is quite a recent development, but I can't pin exactly when it started occurring. Let me know if there is any other useful information I can provide.

WindSoilder:
Sadly found that howdoi *can print newlines* in the code in my environment.  

Would you please tell me that which python version you are running?  I can't reproduce it in `python3`  : (

mulderns:
On my osx, howdoi 1.1.12, python 2.10, python 3.6.4 both show code blocks as single line.

WindSoilder:
After I re-install the `howdoi`, I can reproduce this issue too.

I think this is because the dependency `pyquery` lib have publish a new version `1.4.0`.
Because when I try to install `pyquery` with previous version `1.3.0`, *howdoi* can show code correctly.

BackwardSpy:
I just manually installed `pyquery==1.3.0` in my howdoi venv and it fixed the issue for me too.

gleitz:
This is fixed in #187. Will release a new version shortly

PR #187, opened 2018-02-09
Try to fix for the issue 186

WindSoilder:
Try to fix for the issue #186 
Actually this issue is caused by the recently upgrade of `PyQuery` (which is from 1.3.0 to 1.4.0)
In the new version of `PyQuery`, when we try to extract text from an element, it will *squash html whitespace* by default (which will convert the newline character to a simple space)

To fix this issue, I think there are two solutions:
1.  change *setup.py* file, and feed dependency package with accurate version (like requirements.txt file)
2.  Just use the newer version of `PyQuery`, and make our code fit with that.

Actually, I can't say which solution is better, the solution 1 is more stable than the solution 2.  Because *howdoi* don't need to consider the version of dependency package anymore.

But I think solution 2 can make *howdoi* more active with the dependency package.  If we need to use the new feature of dependency package in the future, we need to update the package and solve the issue again.  And *howdoi* don't have too many dependency packages, I think it's easiler to maintain.

So I take the solution 2 to fix the issue.

And if take solution 2, because the newer version of `PyQuery` don't support python2.6, so remove python2.6 support and update the setup file


gleitz:
Nice!

PR #197, opened 2018-04-22
Display links using markdown syntax

tra38:
From the commit message:

>Some answers on the StackExchange network include hyperlinks
to other content on the Internet. Ideally, those hyperlinks would serve
as "supplemental" reading, but there has been instances where the hyperlink
contains most of the answer (the so-called "link-only answers").

>howdoi currently ignores all hyperlinks within the answer of the
question, preferring instead to display a raw-text version of the answer.
While this is suitable for most instances, there are a few cases
where you may want to dig deeper and follow those hyperlinks, especially
if you have encountered a "link-only answer". Rather than viewing the
StackExchange answer through the browser and risk distraction,
it would be preferable to see the hyperlinks directly in the terminal.

In the future, I may also like to see image URLs show up in the terminal, to handle the case where the answer includes an image as "supplemental" material. Right now, I'm just checking to see if this is a feature that you might want to add to howdoi.

### DIFF
--- a/howdoi/howdoi.py
+++ b/howdoi/howdoi.py
@@ -101,16 +101,14 @@ def _add_links_to_text(element):
     hyperlinks = element.find('a')
 
     for hyperlink in hyperlinks:
+        pquery_object = pq(hyperlink)
         href = hyperlink.attrib['href']
-        copy = hyperlink.text
+        copy = pquery_object.text()
         if (copy == href):
             replacement = copy
         else:
             replacement = "[{0}]({1})".format(copy, href)
-        _replace_link(hyperlink, replacement)
-
-def _replace_link(hyperlink, replacement):
-    pq(hyperlink).replace_with(replacement)
+        pquery_object.replace_with(replacement)
 
 def _get_text(element):
     ''' return inner text in pyquery element '''
--- a/howdoi/howdoi.py
+++ b/howdoi/howdoi.py
@@ -97,8 +97,24 @@ def _get_result(url):
         raise e
 
 
+def _add_links_to_text(element):
+    duplicate_element = element.clone()
+    hyperlinks = duplicate_element.find('a')
+
+    for hyperlink in hyperlinks:
+        href = hyperlink.attrib['href']
+        copy = hyperlink.text
+        markdown = "[{0}]({1})".format(copy, href)
+        _replace_first_link_with_markdown(duplicate_element, markdown)
+
+    return duplicate_element
+
+def _replace_first_link_with_markdown(element, markdown):
+    element.find('a').eq(0).replace_with(markdown)
+
 def _get_text(element):
     ''' return inner text in pyquery element '''
+    element = _add_links_to_text(element)
     return element.text(squash_space=False)
 
 
@@ -186,6 +202,7 @@ def _get_answer(args, links):
     html = pq(page)
 
     first_answer = html('.answer').eq(0)
+
     instructions = first_answer.find('pre') or first_answer.find('code')
     args['tags'] = [t.text for t in html('.post-tag')]
 
--- a/howdoi/howdoi.py
+++ b/howdoi/howdoi.py
@@ -104,13 +104,16 @@ def _add_links_to_text(element):
     for hyperlink in hyperlinks:
         href = hyperlink.attrib['href']
         copy = hyperlink.text
-        markdown = "[{0}]({1})".format(copy, href)
-        _replace_first_link_with_markdown(duplicate_element, markdown)
+        if (copy == href):
+            replacement = copy
+        else:
+            replacement = "[{0}]({1})".format(copy, href)
+        _replace_first_link(duplicate_element, replacement)
 
     return duplicate_element
 
-def _replace_first_link_with_markdown(element, markdown):
-    element.find('a').eq(0).replace_with(markdown)
+def _replace_first_link(element, replacement):
+    element.find('a').eq(0).replace_with(replacement)
 
 def _get_text(element):
     ''' return inner text in pyquery element '''
--- a/howdoi/howdoi.py
+++ b/howdoi/howdoi.py
@@ -98,8 +98,7 @@ def _get_result(url):
 
 
 def _add_links_to_text(element):
-    duplicate_element = element.clone()
-    hyperlinks = duplicate_element.find('a')
+    hyperlinks = element.find('a')
 
     for hyperlink in hyperlinks:
         href = hyperlink.attrib['href']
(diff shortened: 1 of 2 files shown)


tra38:
While doing some manual testing, I realized that converting every hyperlink into markdown will sometimes lead to URLs being duplicated (like ```[https://example.com](https://example.com)```), so I pushed up a new commit that checks first if the "href" and the "copy" are equal to each other. If they are, just return the plain text version.

The automated tests seem to be failing only for Python 2.7...and I'll try to investigate why that is the case and will push up a fix. Here's the output from Travis CI.

>SyntaxError: Non-ASCII character '\xe2' in file /home/travis/build/gleitz/howdoi/test_howdoi.py on line 111, but no encoding declared; see http://python.org/dev/peps/pep-0263/ for details

gleitz:
This is interesting. As I believe most folks use howdoi from the command line, what changes can they expect to see? Are there any specific examples of queries that this handles in a better manner?

tra38:
### Current Behavior
```
>howdoi -a change the position of links within div-header

★  Answer from https://stackoverflow.com/questions/14883784/how-to-change
-the-position-of-these-links-within-div-header ★
In the header, you need to float the h1 left, and the ul right, and then
add display: inline to your links (to keep them on the same line).
To target the list items in the header, you can simply use a parent selector,
like this: #header li.

Here's what you need:
#header {
    border: 1px solid #ccc;
    overflow: hidden; /* clearfix */
}
#header h1 {
    float: left;
}
#header ul {
    float: right;
}
#header li {
    display: inline;
}

See DEMO.
```

```
>howdoi -a difference between c# and java generics

★  Answer from https://stackoverflow.com/questions/355060/c-sharp-vs-java-generics ★
streloksi's link does a great job of breaking down the differences.  The quick and
 dirty summary though is ...
In terms of syntax and usage.  The syntax is roughly the same between the languages. 
 A few quirks here and there (most notably in constraints).  But basically if you can
 read one, you can likely read/use the other.  
The biggest difference though is in the implementation.  
Java uses the notion of type erasure to implement generics.  In short the
 underlying compiled classes are not actually generic.  They compile down to
 Object and casts.  In effect Java generics are a compile time artifact and can
 easily be subverted at runtime.  
C# on the other hand, by virtue of the CLR, implement generics all they way
 down to the byte code.  The CLR took several breaking changes in order to
 support generics in 2.0.  The benefits are performance improvements,
 deep type safety verification and reflection.  
Again the provided link has a much more in depth breakdown
 I encourage you to read
```

### Proposed Behavior
```
>python3 -m howdoi.howdoi -a change the position of links within div-header

★  Answer from https://stackoverflow.com/questions/14883784/how-to-change
-the-position-of-these-links-within-div-header ★
In the header, you need to float the h1 left, and the ul right, and then
add display: inline to your links (to keep them on the same line).
To target the list items in the header, you can simply use a parent selector,
like this: #header li.

Here's what you need:
#header {
    border: 1px solid #ccc;
    overflow: hidden; /* clearfix */
}
#header h1 {
    float: left;
}
#header ul {
    float: right;
}
#header li {
    display: inline;
}

See [DEMO](http://jsfiddle.net/zaknotzak/hYyAZ/).
```

```
>python3 -m howdoi.howdoi -a difference between c# and java generics

★  Answer from https://stackoverflow.com/questions/355060/c-sharp-vs-java-generics ★
[streloksi's link](http://www.jprl.com/Blog/archive/development/2007/Aug-31.html) does a
great job of breaking down the differences.  The quick and dirty summary though is ...
In terms of syntax and usage.  The syntax is roughly the same between the languages.
A few quirks here and there (most notably in constraints).  But basically if you can
 read one, you can likely read/use the other.  
The biggest difference though is in the implementation.  
Java uses the notion of type erasure to implement generics.  In short the underlying
 compiled classes are not actually generic.  They compile down to Object and casts.
 In effect Java generics are a compile time artifact and can easily be subverted at
 runtime.  
C# on the other hand, by virtue of the CLR, implement generics all they way down to the
 byte code.  The CLR took several breaking changes in order to support generics in 2.0. 
 The benefits are performance improvements, deep type safety verification and reflection.  
Again the provided [link](http://www.jprl.com/Blog/archive/development/2007/Aug-31.html)
 has a much more in depth breakdown I encourage you to read
```

howdoi/howdoi.py
```diff
@@ -97,8 +97,27 @@ def _get_result(url):
...
+    hyperlinks = duplicate_element.find('a')
+
+    for hyperlink in hyperlinks:
+        href = hyperlink.attrib['href']
+        copy = hyperlink.text
+        if (copy == href):
+            replacement = copy
+        else:
+            replacement = "[{0}]({1})".format(copy, href)
+        _replace_first_link(duplicate_element, replacement)
```
gleitz: This is clever. I was trying to think of a case where it would fail and can't seem to come up with anything. However, is it possible to simply to call `.replace_with` on the `hyperlink` object within the `for` loop rather than assuming that it will be the first `a` found on the parent element?

test_howdoi.py
```diff
@@ -108,6 +107,41 @@ def test_colorize(self):
         self.assertTrue(normal.find('[39;') is -1)
         self.assertTrue(colorized.find('[39;') is not -1)
 
+    def test_get_text_without_links(self):
```
gleitz: Thanks for adding tests.

howdoi/howdoi.py
tra38: You can only call ```.replace_with``` as a pquery object. In order to get the copy and the ```href``` of the ```hyperlink```, I have to unwrap the pquery object, thereby losing access to the ```.replace_with``` method. When I refactored the code, I "wrapped" the hyperlink back up into a pquery object, so I could then call ```.replace_with``` directly on the element. Still hacky, but less hacky than the old method.

tra38:
While using this modified copy of howdoi (ironically while trying to refactor this PR), I found a bug where the copy is not correctly gathered if it is nested within another tag, like a ```<code>``` tag.

Input HTML:
```
If the function is from a source file available on the filesystem, then <a
href="https://docs.python.org/3/library/inspect.html#inspect.getsource" rel="noreferrer">
<code>inspect.getsource(foo)</code></a> might be of help:
```

Output Text:
```
If the function is from a source file available on the filesystem, then [None]
(https://docs.python.org/3/library/inspect.html#inspect.getsource) might be of help:
```

I'll try to see if I can fix this issue as well.

howdoi/howdoi.py
```diff
@@ -97,8 +97,22 @@ def _get_result(url):
...
+        pquery_object = pq(hyperlink)
+        href = hyperlink.attrib['href']
+        copy = pquery_object.text()
+        if (copy == href):
+            replacement = copy
+        else:
+            replacement = "[{0}]({1})".format(copy, href)
+        pquery_object.replace_with(replacement)
+
 def _get_text(element):
```
WindSoilder: I think the *_get_text* can rename to *get_text*..
Because unit test doesn't mean to test for private function.  And a function which is begin with a single underline is considered as a private method or function.

gleitz: Yes let's call it `get_text`.

PR #218, opened 2018-11-06
Yoba

azhitnik:
This is my yoba branch


ddci:
Looks like this PR will have problems with performence. Please review this code again.

PR #229, opened 2019-04-05
Lazily load pygments to cut start-up time significantly

paulie-g:
Pygments is notorious for slow imports. Load it lazily = win.

### DIFF
--- a/howdoi/howdoi.py
+++ b/howdoi/howdoi.py
@@ -18,6 +18,9 @@ import requests_cache
 import sys
 from . import __version__
 
+from lazyasd import load_module_in_background
+load_module_in_background('pygments')
+
 from pygments import highlight
 from pygments.lexers import guess_lexer, get_lexer_by_name
 from pygments.formatters.terminal import TerminalFormatter
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,3 +5,4 @@ lxml==4.2.5
 pyquery==1.4.0
 requests==2.20.1
 requests-cache==0.4.13
+lazyasd==0.1.4


gleitz:
Paulie G with the zip zip! 🎯

run
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/python
```
gleitz: How important is this `run` file, and when might I use it?

paulie-g: Git acceident, squashed.

gleitz:
I merged and the Travis tests fail on Python 3. Could be unrelated, but any idea of there is a different incantation we need for Python 3?

https://travis-ci.org/gleitz/howdoi/builds/515984660

paulie-g:
Was it passing before the merge?

I tested the change on 3.7.3, so we know the code isn't *actually* broken. Looking at the travis output, it seems the linter is complaining about a few things, then travis just hangs. I wouldn't have thought a few style warnings from flake8 could do that. Curious.

gleitz:
Yes definitely strange. I haven't run those tests in a while. Will check it a little later. 

Thanks for the PR and the ideas on caching.

paulie-g:
Yep, it used to pass, but now it is hanging on nosetests. Maybe it's a Travis problem, or an intermittent error? I don't have commit, but you should have a button somewhere to run Travis again. The unit tests in the repo itself pass on my machine. I can't even see any reference to nosetests in the tree.

paulie-g:
It may be the case that something broke in the Travis env wrt nosetests since you last ran a build.. Not sure I have any ideas beyond that.

paulie-g:
So, I couldn't leave this be and I think I've tracked it down.

strace -f nosetests shows the background thread - the one responsible for loading pygments - blocking on a futex. Essentially, this is a deadlock as the main thread is blocking on attribute access to the module at the same time presumably while holding the GIL. I think this doesn't happen outside of the test suite because as soon as the main thread blocks on network i/o, GIL is released and the import completes.

Code works in production. It does break tests (albeit only when running with nosetests, not directly - I still don't know why). The options are to either leave the patch in and fix the deadlock in future, or back out the patch and, at some point, I'll try to resubmit with a fix (likely in both lazyasd *and* howdoi).

I advise the latter. Besides the formal 'back it out if it breaks tests even if it doesn't break production', we don't know that other people don't use howdoi as an import - if they are, they could well run into the same problem (given I still don't understand why nosetests triggers it).

paulie-g:
Alternatively, we could just do lazy loading conditionally, if not in the testing context. It would kludge nosetests and travis would pass.

gleitz:
Good writeup of the issue. I've reverted in https://github.com/gleitz/howdoi/commit/af99c8265baed111e1dd75b1af28a2205c1bef5b but please re-open when we get an update from lazyasd.

paulie-g:
We could in theory ship with my fork (there's a way to specify github as the source of a package in requirements.txt). I'm not convinced it's a good idea currently. I've done some more profiling on the start-up times and there are bigger fish to fry. The biggest ones are getting rid of the pkg_resources import and the caching change. requests_cache is elegant but far too heavy, especially since it insists on loading its mongodb backend at tremendous cost. When I get a bit of time, expect a few more PRs ;)

Issue #230, opened 2019-04-11
setuptools-installed howdoi is dog slow


paulie-g:
As promised, I've done some digging. Part of the speed problem is setuptools. Doing a `python setup.py install` causes a pkg_resources import in the launch script. This costs over 200ms on my machine (with a pcie ssd + warm cache). This does *not* happen if you install with a wheel, so either via pip from pypi or `python setup.py bdist_wheel + pip install dist/something.whl`

This may not matter to _you_, me or people who use pip ahead of their distro's package manager, but for everyone else it's a problem since packagers are unlikely to bother with wheels (and the added requirements). The AUR package uses setuptools for example, and I'm sure so do others.

See:
* as is, importing pkg_resources naively (cache hit)
(paul@redacted)[~/dev/howdoi/howdoi]$ hyperfine  'howdoi bash history search'
Benchmark #1: howdoi bash history search
  Time (mean ± σ):     439.9 ms ±   4.0 ms    [User: 383.4 ms, System: 49.5 ms]
  Range (min … max):   435.8 ms … 448.9 ms    10 runs
 
* fast-entry_points magic (cache hit)
(paul@redacted)[~/dev/howdoi/howdoi]$ hyperfine  'howdoi bash history search'
Benchmark #1: howdoi bash history search
  Time (mean ± σ):     221.8 ms ±   3.7 ms    [User: 186.7 ms, System: 28.8 ms]
  Range (min … max):   217.4 ms … 229.3 ms    13 runs

* installed with pip + wheel (cache hit)
(paul@redacted)[~/dev/howdoi/howdoi]$ hyperfine 'howdoi bash history search'
Benchmark #1: howdoi bash history search
  Time (mean ± σ):     214.3 ms ±   4.5 ms    [User: 187.7 ms, System: 26.5 ms]
  Range (min … max):   208.0 ms … 223.5 ms    14 runs

I would argue that given we can't control the packagers, it's best to fix it in howdoi itself.

gleitz:
Thanks for digging. Does your PR address the issue and we can close? Or is there more you would like to do?

paulie-g:
Yes, my PR works around the issue. Wheels are still faster by ~2ms, but it's close. We might want to stick a nota bene in the readme for packagers to use wheels if they can (I can submit a PR for that too if you like).

I do intend to dig further (replacement caching, which is half-way done), but it'll be a separate issue and separate PR.

gleitz:
[...]

PR #231, opened 2019-04-11
Speed up setuptools-installed howdoi by ~50%

paulie-g:
This is some dark magic (well, greyish) to make the launch script generated by setuptools as fast as the one generated when installing a wheel. Makes for 200+ms speed-up (about 50%) on my box. Even bigger for python2. See #230 

I would have preferred to include fastentrypoints as a submodule, but the way that repo is structured it's not worth it for the added complexity. Including it as a requirement would be equally problematic. It is intended to be vendored according to the authors, so that's what we're doing here.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,1 +1,2 @@
 include CHANGES.txt
+include fastentrypoints.py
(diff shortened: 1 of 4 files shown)


gleitz:
Nice!

Issue #224, opened 2018-12-24
Implement simpler caching scheme


Almenon:
As I found out in #213 requests_cache has some severe performance issues. By switching to a simpler caching scheme we should be able to massively speed up start time while still offering caching for the last X requests, X being however many we can store untill running into perf problems on a HDD. 

The caching scheme could be as simple as just appending the request/answer as csv. Or maybe as JSON? Or storing requests in one file and answers in another?  Or using [diskcache](https://github.com/grantjenks/python-diskcache)?  And I'd have to get rid of oldest entries to prevent cache from growing too large too.

Anyways, I'm planning on implementing this, Just wanted to give you a heads up before I start work.

gleitz:
Sounds wild!

paulie-g:
Why not just use an array of N (argv, output) tuples in a pickle? Load it at start and concurrently with the first network request. If it's a hit, output and cancel network request; if it's a miss, after producing output evict one entry from the array by LRU and pickle onto disk. Speed-up for hits, virtually no visible penalties for the vastly more common miss case.

paulie-g:
So the initial commit is in (to fastercache branch in my fork). See: https://github.com/paulie-g/howdoi/commit/24bfdcad98c605e0d172ae6b598bdd49ed2f9d65

requests_cache is removed completely. Instead, we're using cachelib.FileSystemCache (cachelib is the caching stuff from werkzeug/pallets/flask/etc broken out into a separate package). It's not perfect, but it's the best I could find without rolling my own. Basic principle is you give it a directory, key = filename and contents=pickled(date-modified,value). We set it to hold x items (currently 64 iirc) and it automatically purges when it needs to (suboptimally). I'd like to improve it in a number of ways, but it's not clear the loss of maintainability is worth it.

Results are clear, I get ~70ms (down from ~220 to ~150) in the hit case and even a bit more in the miss case. It's only 10-15% of total runtime in the common, miss case on my box, but it's the last of the big wins and very much worth it imo.

With this initial commit, we're caching output directly and keying it on all arguments. This is great when you're just re-running a search. It leaves the case where the user now wishes to colorize (or not), display more answers etc on a subsequent run uncached. I see no use in paying for caching in the common case to serve those cases, but if you do, I'll add a second-level cache to handle them.

[...]

PR #232, opened 2019-04-13
Replace requests_cache with a light, filesystem-based cache

paulie-g:
See #224 for overall discussion.

Additional changes not mentioned there:
 * Use appdirs to find cache directory less kludgily, should now work on Windows as a by-product (not tested though, don't have a Windows machine)
 * Remove import random and use a simple internal randomizer based on os.urandom to select User-Agent

Still to come:
* Second-level cache so we don't have to hit upstream again when user wants full answer, second answer, different colorization setting etc

### DIFF
--- a/howdoi/howdoi.py
+++ b/howdoi/howdoi.py
@@ -91,7 +91,7 @@ else:
 
 howdoi_session = requests.session()
 
-class BlockError(Exception):
+class BlockError(RuntimeError):
     pass
 
 def _random_int(width):
--- a/howdoi/howdoi.py
+++ b/howdoi/howdoi.py
@@ -14,7 +14,7 @@ import argparse
 import os
 import appdirs
 import re
-from cachelib import FileSystemCache
+from cachelib import FileSystemCache, NullCache
 import requests
 import sys
 from . import __version__
@@ -70,10 +70,10 @@ ANSWER_HEADER = u('{2}  Answer from {0} {2}\n{1}')
 NO_ANSWER_MSG = '< no answer given >'
 
 CACHE_DIR = appdirs.user_cache_dir('howdoi')
-CACHE_ENTRY_MAX = 64
+CACHE_ENTRY_MAX = 128
 
 if os.getenv('HOWDOI_DISABLE_CACHE'):
-    cache = False
+    cache = NullCache() # works like an always empty cache, cleaner than 'if cache:' everywhere
 else:
     cache = FileSystemCache(CACHE_DIR, CACHE_ENTRY_MAX, default_timeout=0)
 
@@ -218,7 +218,13 @@ def _get_answer(args, links):
         return False
     if args.get('link'):
         return link
-    page = _get_result(link + '?answertab=votes')
+
+    cache_key = link 
+    page = cache.get(link)
+    if not page:
+        page = _get_result(link + '?answertab=votes')
+        cache.set(cache_key, page)
+
     html = pq(page)
 
     first_answer = html('.answer').eq(0)
@@ -245,13 +251,25 @@ def _get_answer(args, links):
     text = text.strip()
     return text
 
+def _get_cachable_questions(query):
+    cache_key = query + "-links"
+    res = cache.get(cache_key)
+    if res:
+        if res == "NULL":
+            res = False
+        return res
 
-def _get_instructions(args):
-    links = _get_links(args['query'])
+    links = _get_links(query)
     if not links:
-        return False
+        cache.set(cache_key, "NULL")
 
     question_links = _get_questions(links)
+    cache.set(cache_key, question_links or "NULL")
+
+    return question_links
+
+def _get_instructions(args):
+    question_links = _get_cachable_questions(args['query'])
     if not question_links:
         return False
 
@@ -294,17 +312,15 @@ def howdoi(args):
     args['query'] = ' '.join(args['query']).replace('?', '')
     cache_key = str(args)
 
-    if cache:
-        res = cache.get(cache_key)
-        if res:
-            return res
+    res = cache.get(cache_key)
+    if res:
+        return res
 
     try:
         res = _get_instructions(args) 
         if not res:
             res = 'Sorry, couldn\'t find any help with that topic\n'
-        if cache:
-            r = cache.set(cache_key, res)
+        cache.set(cache_key, res)
             
         return res
     except (ConnectionError, SSLError):
--- a/howdoi/howdoi.py
+++ b/howdoi/howdoi.py
@@ -231,7 +231,6 @@ def _get_questions(links):
     return [link for link in links if _is_question(link)]
 
 
-
 def _get_answer(args, links):
     link = get_link_at_pos(links, args['pos'])
     if not link:
@@ -271,7 +270,7 @@ def _get_answer(args, links):
     text = text.strip()
     return text
 
-def _get_cachable_questions(query):
+def _get_links_with_cache(query):
     cache_key = query + "-links"
     res = cache.get(cache_key)
     if res:
@@ -289,7 +288,7 @@ def _get_cachable_questions(query):
     return question_links
 
 def _get_instructions(args):
-    question_links = _get_cachable_questions(args['query'])
+    question_links = _get_links_with_cache(args['query'])
     if not question_links:
         return False
 
--- a/howdoi/howdoi.py
+++ b/howdoi/howdoi.py
@@ -75,6 +75,7 @@ STAR_HEADER = u('\u2605')
 ANSWER_HEADER = u('{2}  Answer from {0} {2}\n{1}')
 NO_ANSWER_MSG = '< no answer given >'
 
+CACHE_EMPTY_VAL = "NULL"
 CACHE_DIR = appdirs.user_cache_dir('howdoi')
 CACHE_ENTRY_MAX = 128
 
@@ -274,16 +275,16 @@ def _get_links_with_cache(query):
     cache_key = query + "-links"
     res = cache.get(cache_key)
     if res:
-        if res == "NULL":
+        if res == CACHE_EMPTY_VAL:
             res = False
         return res
 
     links = _get_links(query)
     if not links:
-        cache.set(cache_key, "NULL")
+        cache.set(cache_key, CACHE_EMPTY_VAL)
 
     question_links = _get_questions(links)
-    cache.set(cache_key, question_links or "NULL")
+    cache.set(cache_key, question_links or CACHE_EMPTY_VAL)
 
     return question_links
 
--- a/howdoi/howdoi.py
+++ b/howdoi/howdoi.py
@@ -13,7 +13,6 @@ gc.disable() # disable right at the start, we don't need it
 import argparse
 import os
 import appdirs
-import random
 import re
 from cachelib import FileSystemCache
 import requests
@@ -80,6 +79,18 @@ else:
 
 howdoi_session = requests.session()
 
+def _random_int(width):
+    bres = os.urandom(width)
+    if sys.version < '3':
+        ires = int(bres.encode('hex'), 16)
+    else:
+        ires = int.from_bytes(bres, 'little')
+
+    return ires
+
+def _random_choice(seq):
+    return seq[_random_int(1) % len(seq)]
+
 def get_proxies():
     proxies = getproxies()
     filtered_proxies = {}
@@ -94,7 +105,7 @@ def get_proxies():
 
 def _get_result(url):
     try:
-        return howdoi_session.get(url, headers={'User-Agent': random.choice(USER_AGENTS)},
+        return howdoi_session.get(url, headers={'User-Agent': _random_choice(USER_AGENTS)},
                                   proxies=get_proxies(),
                                   verify=VERIFY_SSL_CERTIFICATE).text
     except requests.exceptions.SSLError as e:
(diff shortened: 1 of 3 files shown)


paulie-g:
So, we now have three levels of caching, we now detect when google blocked us -> print a helpful message instead of saying there are no answers -> don't cache a not-found result.

howdoi/howdoi.py
```diff
@@ -66,22 +65,40 @@ def u(x):
...
-        os.makedirs(CACHE_DIR)
-    howdoi_session = requests_cache.CachedSession(CACHE_FILE)
+    cache = FileSystemCache(CACHE_DIR, CACHE_ENTRY_MAX, default_timeout=0)
+
+howdoi_session = requests.session()
+
+class BlockError(Exception):
+    pass
+
+def _random_int(width):
```
gleitz: Why the move away from `random`?

howdoi/howdoi.py
```diff
@@ -148,12 +165,23 @@ def _extract_links(html, search_engine):
 def _get_search_url(search_engine):
     return SEARCH_URLS.get(search_engine, SEARCH_URLS['google'])
 
+def _detect_block(page):
```
gleitz: I think I like something more descriptive like `_is_blocked`

howdoi/howdoi.py
```diff
@@ -148,12 +165,23 @@ def _extract_links(html, search_engine):
...
 
 def _get_links(query):
     search_engine = os.getenv('HOWDOI_SEARCH_ENGINE', 'google')
     search_url = _get_search_url(search_engine)
 
     result = _get_result(search_url.format(URL, url_quote(query)))
+    if _detect_block(result):
+        print('[ERROR] Encountered a temporary block due to automated request '
+                'detection. Please wait a few minutes for it to clear or change search engine')
+        raise BlockError("Temporary block by search engine")
```
gleitz: What about using a standard exception like https://docs.python.org/3/library/exceptions.html#RuntimeError

howdoi/howdoi.py
```diff
@@ -203,13 +231,20 @@ def _get_questions(links):
     return [link for link in links if _is_question(link)]
 
 
+
```
gleitz: No need for an extra line here

howdoi/howdoi.py
```diff
@@ -66,22 +65,40 @@ def u(x):
...
 ANSWER_HEADER = u('{2}  Answer from {0} {2}\n{1}')
 NO_ANSWER_MSG = '< no answer given >'
-XDG_CACHE_DIR = os.environ.get('XDG_CACHE_HOME',
-                               os.path.join(os.path.expanduser('~'), '.cache'))
-CACHE_DIR = os.path.join(XDG_CACHE_DIR, 'howdoi')
-CACHE_FILE = os.path.join(CACHE_DIR, 'cache{0}'.format(
-    sys.version_info[0] if sys.version_info[0] == 3 else ''))
+
+CACHE_DIR = appdirs.user_cache_dir('howdoi')
+CACHE_ENTRY_MAX = 128
```
gleitz: Should this be configurable through `os.environ.get`?

howdoi/howdoi.py
```diff
@@ -66,22 +65,40 @@ def u(x):
...
 
 if os.getenv('HOWDOI_DISABLE_CACHE'):
-    howdoi_session = requests.session()
+    cache = NullCache() # works like an always empty cache, cleaner than 'if cache:' everywhere
 else:
-    if not os.path.exists(CACHE_DIR):
-        os.makedirs(CACHE_DIR)
-    howdoi_session = requests_cache.CachedSession(CACHE_FILE)
+    cache = FileSystemCache(CACHE_DIR, CACHE_ENTRY_MAX, default_timeout=0)
```
gleitz: Do we want to allow for cache expiration after some amount of time? Perhaps it's 0 by default, but eventually expires since StackOverflow answers are constantly changing and getting better.

howdoi/howdoi.py
```diff
@@ -236,13 +271,25 @@ def _get_answer(args, links):
     text = text.strip()
     return text
 
+def _get_cachable_questions(query):
```
gleitz: Since this ultimately returns links, perhaps `_get_links_with_cache()` makes more sense.

howdoi/howdoi.py
```diff
@@ -236,13 +271,25 @@ def _get_answer(args, links):
...
+    cache_key = query + "-links"
+    res = cache.get(cache_key)
+    if res:
+        if res == "NULL":
```
gleitz: `"NULL"` feels bit gross to me. Is it possible to put `None` directly into the cache? If not, then let's put this as `CACHE_EMPTY_PLACEHOLDER = "NULL"` or similar at the top and use it instead of retyping the string.

howdoi/howdoi.py
```diff
@@ -315,8 +375,11 @@ def command_line_runner():
         return
 
     if args['clear_cache']:
-        _clear_cache()
-        print('Cache cleared successfully')
+        if _clear_cache():
+            print('Cache cleared successfully')
+        else:
+            print('Error: clearing cache failed')
```
gleitz: Was thinking about the way we signal errors to the user. We now have `[ERROR]`, `Error:`, or simply the message itself like`'Sorry, couldn\'t find any help with that topic\n'`. I think I like simply saying that `Unable to clear the cache`.

howdoi/howdoi.py
```diff
@@ -148,12 +165,23 @@ def _extract_links(html, search_engine):
...
+
+    return False
 
 def _get_links(query):
     search_engine = os.getenv('HOWDOI_SEARCH_ENGINE', 'google')
     search_url = _get_search_url(search_engine)
 
     result = _get_result(search_url.format(URL, url_quote(query)))
+    if _detect_block(result):
+        print('[ERROR] Encountered a temporary block due to automated request '
```
gleitz: How about:

Unable to find an answer because the search engine blocked the request. Please wait a few minutes or select a different search engine.

howdoi/howdoi.py
paulie-g: Small win. Easy to do, minimal code, 0 portability issues.

howdoi/howdoi.py
paulie-g: k

howdoi/howdoi.py
paulie-g: Was considered good practice on other py projects I contributed to. I happen to like it and it makes for cleaner code if we end up handling the exception upstream instead of just oopsing with a tb. Your call, I don't insist.

howdoi/howdoi.py
paulie-g: k

howdoi/howdoi.py
paulie-g: Imo, no. I wouldn't make it any larger (the lib has some O(n) behaviour w disk access on prune), and I wouldn't make it much smaller or it won't be useful. Making a useful change requires understanding how the cache works. If I ever get around to forking+fixing/improving cachelib, we can expose it.

howdoi/howdoi.py
paulie-g: Yes, there should be one function that does that which has one format for standard messages and one for errors. It's not cache-related though and I wanted to avoid piling things in.

paulie-g: With that said, it's an error and should be clearly indicated as such. If there are ever tests, they'll want it too.

howdoi/howdoi.py
paulie-g: Disagree. It's an error and should be indicated as such.

howdoi/howdoi.py
paulie-g: Keep in mind that it's a rolling cache with 128 entries. Each new invocation creates at least 3 + num_answers. I don't expect the cache to get stale. With that said, it's easy to set; we can even set different TTLs on output/links and answer-html entries (there is a timeout option on cache.set()). If you come up with values, I'll put them in.

howdoi/howdoi.py
paulie-g: You can put it in, but you then have no way to distinguish from the key not being present (cache.get() returns None). This is something a good api would account for, but we don't have one. A special Empty class that's bool(res)=true and behaves like None otherwise would be elegant. I don't see the point, but it's your call.

paulie-g:
We're in danger of spending too much time debating minutiae. It's your project, so your call on all of the cosmetic issues. You want it done, let me know and it shall be so.

howdoi/howdoi.py
gleitz: Didn't realize it was rolling. This works.

howdoi/howdoi.py
gleitz: OK then let's do the `CACHE_EMPTY_PLACEHOLDER = "NULL"` at the top

howdoi/howdoi.py
gleitz: I agree that errors should be marked. Let's just pick one style `[ERROR]` or `Error:`

howdoi/howdoi.py
paulie-g: NULL is a well-known constant used from C to YAML and pickles to something short. I don't see it as any worse than def class EmptyVal -> isinstance(res, EmptyVal), but I don't want to gross you out ;)

howdoi/howdoi.py
gleitz: If you want to include the error statement, then let's go with `[ERROR] Unable to find an answer because the search engine temporarily blocked the request Please wait a few minutes or select a different search engine.`

howdoi/howdoi.py
gleitz: OK let's keep the BlockError, but it should inherit from RuntimeError, yes?

howdoi/howdoi.py
paulie-g: Yes, it's LIFO pruned to CACHE_ENTRY_MAX. The pruning is actually what is O(n) - and what I would fix first (they pickle modification date, so have to open file to check; it should be done with stat() instead, with dir still in cache it wouldn't hit disk).

howdoi/howdoi.py
paulie-g: It replaces calls to _get_links + _get_questions in case of cache hit, so it's a direct rename. _get_links returns links, _get_questions returns questions ;)

howdoi/howdoi.py
gleitz: I don't think that generating random numbers or selecting random elements should be re-written in this repo. In a bit a research I did see that `random` isn't cryptographically secure and thus you might want to use `urandom`, but unless there are serious portability or speed considerations with `random` then let's use the library.

howdoi/howdoi.py
paulie-g: If you're asking me, no - RuntimeError is for people too lazy do define their own exceptions properly, says nothing and gives no benefit. I have exactly 0 problem just doing what you ask though :)

paulie-g:
I'll be afk for an hour or so. When I come back, I'll take stock of what we have here at this point and make all the changes you are still requesting.

howdoi/howdoi.py
gleitz: I'm suggesting the following:

```
class BlockError(RuntimeError):
    pass
```

Here, we get the advantages of a custom error, but we're more explicit about the type of Exception we're handling – one that happens at runtime and thus out of the hands of the developer.

howdoi/howdoi.py
gleitz: Yes. what I want to avoid is the string "NULL" in more than one place in the code.

howdoi/howdoi.py
paulie-g: Yes, there's a few ms win here from removing the import, more on a non-ssd machine with a cold cache. It's what, 10 lines of perfectly straightforward code? I went through the whole import profile and looked if anything could be jettisoned without much trouble. This was the easiest one. It's bad to re-implement for no reason, but I'm lazy and always have a reason if I'm going to do something ;)

howdoi/howdoi.py
paulie-g: 6abb8420a67cf0a65abc91fcd5b9462735aaa286

howdoi/howdoi.py
paulie-g: 6abb8420a67cf0a65abc91fcd5b9462735aaa286

howdoi/howdoi.py
paulie-g: 4c56e752d3150eab8f8537241877151b8bb59ebd

howdoi/howdoi.py
paulie-g: ab18e0df0219c68f689731667e8a26966029414c

howdoi/howdoi.py
paulie-g: ab18e0df0219c68f689731667e8a26966029414c

howdoi/howdoi.py
paulie-g: I understood what you were suggesting. All exceptions defined in running code are by definition runtime exceptions. I consider inheriting from built-in exceptions unidiomatic (unless it's a clear sub-type) and haven't seen it in Twisted or any other large projects I'm familiar with - they all have their own exception hierarchies. With that said, it's not important and deffo your call.

d3e69832005107745342e04317597ff2fe0aa639

howdoi/howdoi.py
paulie-g: 529a6727b4f5961a3b99ea79e115f3bee86d497a

howdoi/howdoi.py
paulie-g: It's essentially the equivalent of inlining something for perf purposes in C. I'll remove it if you absolutely insist, but I don't find the arguments against doing this compelling.

gleitz:
Thx @paulie-g!

PR #235, opened 2019-04-15
More granular control over HOWDOI_SEARCH_ENGINE

rtruxal:
added a bing arg to the cmdline interface so HOWDOI_SEARCH_ENGINE can be set within the howdoi process itself.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -100,7 +100,7 @@ Usage
                             number of answers to return
       -C, --clear-cache     clear the cache
       -v, --version         displays the current version of howdoi
-      -e [ENGINE], --search-engine [ENGINE]  change search engine for this query only
+      -e ENGINE, --search-engine ENGINE  change search engine for this query only
 
 
 As a shortcut, if you commonly use the same parameters each time and don't want to type them, add something similar to your .bash_profile (or otherwise). This example gives you 5 colored results each time.
--- a/README.rst
+++ b/README.rst
@@ -83,7 +83,7 @@ Usage
 
 ::
 
-    usage: howdoi.py [-h] [-p POS] [-a] [-l] [-c] [-n NUM_ANSWERS] [-C] [-v] QUERY [QUERY ...]
+    usage: howdoi.py [-h] [-p POS] [-a] [-l] [-c] [-n NUM_ANSWERS] [-C] [-v] [-e ENGINE] QUERY [QUERY ...]
 
     instant coding answers via the command line
 
@@ -100,7 +100,7 @@ Usage
                             number of answers to return
       -C, --clear-cache     clear the cache
       -v, --version         displays the current version of howdoi
-      -b, --bing            change search engine to bing for this query only
+      -e [ENGINE], --search-engine [ENGINE]  change search engine for this query only
 
 
 As a shortcut, if you commonly use the same parameters each time and don't want to type them, add something similar to your .bash_profile (or otherwise). This example gives you 5 colored results each time.
--- a/README.rst
+++ b/README.rst
@@ -130,7 +130,7 @@ Notes
 -  Slack integration available through `slack-howdoi <https://github.com/ellisonleao/slack-howdoi>`_.
 -  Howdoi uses a cache for faster access to previous questions. Caching functionality can be disabled by setting the HOWDOI_DISABLE_CACHE environment variable. The cache is stored in `~/.cache/howdoi`.
 -  You can set the HOWDOI_URL environment variable to change the source url for answers (default: `stackoverflow.com`, also supported: `serverfault.com`, `pt.stackoverflow.com`, `full list <http://stackexchange.com/sites?view=list#traffic>`_).
--  You can set the HOWDOI_SEARCH_ENGINE environment variable to change the underlying search engine for StackOverflow links (default: `google`, also supported: `bing`). The -b flag will switch the underlying engine to bing for a single query.
+-  You can set the HOWDOI_SEARCH_ENGINE environment variable to change the underlying search engine for StackOverflow links (default: `google`, also supported: `bing`). The -e flag will switch the underlying engine for a single query.
 -  Setting the HOWDOI_COLORIZE environment variable will colorize the output by default.
 -  Special thanks to Rich Jones (`@miserlou <https://github.com/miserlou>`_) for the idea.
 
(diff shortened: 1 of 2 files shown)


WindSoilder:
If we want to add a command line arguments, I would recommend using `-e --engine bing`  rather than simple `-b` argument.

If we use simple `-b` argument, what if we need to support another search engine like `Duckduckgo` or `yandex`?  Then we will mess up when we support more search engines in the user interface.

rtruxal:
Did it.

paulie-g:
This is going to need some work with the new caching code in the fastcache branch.

gleitz:
I'm wrestling with adding additional command line arguments that can already be achieved with environment variables. For example consider the following:

```
alias howdoi_bing='HOWDOI_SEARCH_ENGINE="bing" howdoi'
```

then

```
» howdoi -l -n3 format date bash
https://stackoverflow.com/questions/1401482/yyyy-mm-dd-format-date-in-shell-script

================================================================================

https://webcache.googleusercontent.com/search?q=cache:Oar9zh1xk10J:https://stackoverflow.com/questions/1401482/yyyy-mm-dd-format-date-in-shell-script+&cd=1&hl=en&ct=clnk&gl=us

================================================================================

https://stackoverflow.com/questions/6508819/convert-date-formats-in-bash
```

is different than 

```
» howdoi_bing -l -n3 format date bash
https://stackoverflow.com/questions/1401482/yyyy-mm-dd-format-date-in-shell-script

================================================================================

https://stackoverflow.com/questions/6508819/convert-date-formats-in-bash

================================================================================

https://stackoverflow.com/questions/14152759/linux-bash-date-format
```

Thoughts?

paulie-g:
You're absolutely right. With that said, environment variables ought not be the primary interface - that's just poor UX. However, bing results are so poor for howdoi's purposes that I don't think it matters - no one other than people who can't use google for some reason are likely to want to use bing.

gleitz:
Yes the main reason I could see users switching away from Google is if they're blocked due to a rate limit, or if using howdoi from a country that blocks Google. I think the best thing to do would be a note on the README page that includes the following snippet

```
alias howdoi_bing='HOWDOI_SEARCH_ENGINE="bing" howdoi'
```

paulie-g:
I'm not opposed to adding this in future, I just don't currently have the bandwidth to go through what this would do to the new caching. Probably nothing bad other than output cache not hitting, I just don't recall (several ksloc deep into a couple other projects atm).

In principle, I'd like to see a better config layer that accepts a .py config file in appdirs.user_config_dir, env vars, command line arguments or py dict when howdoi is used as a library.  And a man page. I'd take that over ad hoc tweaks to particular settings.

paulie-g:
To be clear, I see no harm in adding this and no reason not to do it, provided it doesn't break anything. I'd say wait until dev is merged into master, rebase/eliminate merge conflicts if necessary, and wait for @rtruxal to test and confirm that it doesn't break the new caching (by break I mean errors, not misses). Being picky with relatively small patches that are clearly not net negative is not a good way to keep new contributors involved and interested imo.

gleitz:
Ok I just merged dev into master. Would you please rebase @rtruxal?

README.rst
```diff
@@ -129,7 +130,7 @@ Notes
 -  Slack integration available through `slack-howdoi <https://github.com/ellisonleao/slack-howdoi>`_.
 -  Howdoi uses a cache for faster access to previous questions. Caching functionality can be disabled by setting the HOWDOI_DISABLE_CACHE environment variable. The cache is stored in `~/.cache/howdoi`.
 -  You can set the HOWDOI_URL environment variable to change the source url for answers (default: `stackoverflow.com`, also supported: `serverfault.com`, `pt.stackoverflow.com`, `full list <http://stackexchange.com/sites?view=list#traffic>`_).
--  You can set the HOWDOI_SEARCH_ENGINE environment variable to change the underlying search engine for StackOverflow links (default: `google`, also supported: `bing`).
+-  You can set the HOWDOI_SEARCH_ENGINE environment variable to change the underlying search engine for StackOverflow links (default: `google`, also supported: `bing`). The -b flag will switch the underlying engine to bing for a single query.
```
paulie-g: This is still referencing the flag as -b, rather than -e.

paulie-g:
Appears to have no conflicts. @rtruxal please test that it doesn't break caching (run with HOWDOI_DISABLE_CACHE, without, on clear/full cache and so on), and address https://github.com/gleitz/howdoi/pull/235/files#r278707922

If all goes well, and by @gleitz 's assent, I think this can go in :)

rtruxal:
@paulie-g - Sounds good. I will try to do this over the weekend.

gleitz:
@rtruxal no rush but lemme know if those checks passes and you're ready to merge.

rtruxal:
Gahhh thank you for the reminder.

rtruxal:
So good-news/bad-news:

good news - I checked the new arg against the `fastercache` branch and everything appears to be fine (using `HOWDOI_DISABLE_CACHE`, full/empty cache)

bad-news - The optional positional arg doesn't work. `howdoi -e do things` gets broken up into `howdoi ==> -e do ==> things`. I can't immediately think of a way to fix this, other than switching it back or making the positional argument mandatory.

gleitz:
I think the correct usage should be

`howdoi -e bing do things`

README.rst
```diff
@@ -100,6 +100,7 @@ Usage
                             number of answers to return
       -C, --clear-cache     clear the cache
       -v, --version         displays the current version of howdoi
+      -e [ENGINE], --search-engine [ENGINE]  change search engine for this query only
```
gleitz: Here I would list the possible engines (google, bing)

gleitz: Also `--engine` is fine for the long name

howdoi/howdoi.py
```diff
@@ -303,6 +306,8 @@ def get_parser():
                         action='store_true')
     parser.add_argument('-v', '--version', help='displays the current version of howdoi',
                         action='store_true')
+    parser.add_argument('-e', '--search-engine', help='change search engine for this query only',
```
gleitz: Let's use `--engine`

rtruxal: https://github.com/gleitz/howdoi/pull/235/commits/4c4219456c50daddc732d2c6bafe417a0acb76bb

README.rst
rtruxal: https://github.com/gleitz/howdoi/pull/235/commits/c1df93fc892761377698738d6089b72b3949b076

gleitz:
Thanks for this!

PR #240, opened 2019-05-20
Fix inconsistency between command line parser definition and use of t…

hdboer:
…he 'search_engine' key in command_line_runner() which causes a KeyError exception.


gleitz:
Thanks! Happy hacking

PR #243, opened 2019-06-25
Explicitly set English interface in SEARCH_URLS

Subc2:
In my country Google adds a "Translate site" link to most of the results in foreign languages, which breaks howdoi. The `hl` parameter fixes the issue without breaking other search engines, including Bing.

### DIFF
--- a/howdoi/howdoi.py
+++ b/howdoi/howdoi.py
@@ -72,8 +72,8 @@ USER_AGENTS = ('Mozilla/5.0 (Macintosh; Intel Mac OS X 10.7; rv:11.0) Gecko/2010
                ('Mozilla/5.0 (Windows; Windows NT 6.1) AppleWebKit/536.5 (KHTML, like Gecko) Chrome/19.0.1084.46'
                 'Safari/536.5'), )
 SEARCH_URLS = {
-    'bing': SCHEME + 'www.bing.com/search?q=site:{0}%20{1}',
-    'google': SCHEME + 'www.google.com/search?q=site:{0}%20{1}'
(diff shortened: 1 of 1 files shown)


gleitz:
Thanks for this. Does this always force English results? I know some users like to set the HOWDOI_URL to something else such as pt.stackoverflow.com. I worry that this addition, in conjunction with that URL, might not give the best results.

Subc2:
The `host language` parameter indeed affects the search results. It seems that Google "treats" results in the interface language as more important. However, the impact of this option is greatly reduced when searching a domain written in only one language, as is the case with Stack Overflow variants.

As a test I searched `ru.stackoverflow.com` for a deliberately vague phrase, translating to "how to call a terminal":
- https://www.google.com/search?q=%D0%9A%D0%B0%D0%BA+%D0%B2%D1%8B%D0%B7%D0%B2%D0%B0%D1%82%D1%8C+%D1%82%D0%B5%D1%80%D0%BC%D0%B8%D0%BD%D0%B0%D0%BB+site%3Aru.stackoverflow.com&hl=ru
- https://www.google.com/search?q=%D0%9A%D0%B0%D0%BA+%D0%B2%D1%8B%D0%B7%D0%B2%D0%B0%D1%82%D1%8C+%D1%82%D0%B5%D1%80%D0%BC%D0%B8%D0%BD%D0%B0%D0%BB+site%3Aru.stackoverflow.com&hl=en

The top 8 results were the same when using `hl=ru` and `hl=en` (one result changed its place).

Therefore, I conclude that introducing this option (for now) causes little to no harm.

gleitz:
Excellent. Thanks for checking.

PR #246, opened 2019-08-08
Implement duck duck go search engine

marek2901:


gleitz:
With passing tests. Very nice!

PR #253, opened 2019-09-21
added blocked detection for duckduckgo

claudioacioli:


test_howdoi.py
```diff
@@ -68,14 +68,17 @@ def test_answers_bing(self):
...
+            for query in self.queries:
+                self.assertTrue(self.call_howdoi(query))
+            for query in self.bad_queries:
+                self.assertTrue(self.call_howdoi(query))
+
+            os.environ['HOWDOI_URL'] = 'pt.stackoverflow.com'
+            for query in self.pt_queries:
+                self.assertTrue(self.call_howdoi(query))
+        except howdoi.BlockError:
+            pass
```
gleitz: This is clever, but I worry that if we're getting blocked the tests shouldn't just pass. Maybe we fail in this case with an error message like "Blocked by search engine"

claudioacioli: Thank you for your feedback. I will try do this and resend pull request. Sorry, i need learning more about test cases.

gleitz: No worries. We're not actually using best practices with testing. Ideally it wouldn't make a live call - we would instead mock the responses from Goog, Bing, Duck and use those instead.

claudioacioli: Yes, I opened and closed a pull request, #252, the live call on Duckduckgo produced fails of tests of Travis, so I changed the test file.

What do you think about these solutions?

```
#1
except howdoi.BlockError as e:
    print "\nBlockError Exception: %s" % str(e)
````

```
#2
except howdoi.BlockError as e:
    self.assertEqual(e.value, "Temporary block by search engine")
````

Comment (gleitz) [reply test_howdoi.py]: Let's got with #1, logging that there was a block exception:
```
_print_err('Unable to complete test due to BlockError')
```

gleitz: We can go ahead and let the test pass for now. Maybe in the future we can rewrite the tests.

claudioacioli: If you want, accept this pull request and to the night I will send the code with the solution.

Thanks a lot for your time!

PR #257, opened 2019-09-25
added blocked detection for duckduckgo

claudioacioli:


claudioacioli:
I really don't know if this is the correct way.

PR #260, opened 2020-02-02
Move function to correct position

RFV5s:
the function is now declared before it is used


gleitz:
For some reason the nose tests fail in Python 2.7
https://travis-ci.org/gleitz/howdoi/builds/645016396

PR #272, opened 2020-06-12
Auto parse args

gmelodie:
`howdoi.howdoi` accepts both parsed and raw queries to be passed directly to it. So the following snippet:
```python
from howdoi import howdoi

query = "for loop python"
parser = howdoi.get_parser()
args = vars(parser.parse_args(query.split(' ')))

output = howdoi.howdoi(args)
```
Can **also** be written as:
```python
from howdoi import howdoi

query = "for loop python"
output = howdoi.howdoi(query)
```
This doesn't break compatibility with the previous version.


gmelodie:
Removed undesired commits on #273

PR #276, opened 2020-06-15
Add json support

### LINKED ISSUE #274: JSON response format
Be able to return JSON from howdoi, perhaps with the `-j / --json` flag

gmelodie:
Add support for json responses (all data is handled in JSON internally)

Fixes #274 and part of #278


chrisngyn:
@YKo20010 and I decided to add a new `-jf` flag. I recall in our last meeting we talked about how the functionality of some of the features the others were working on used the raw JSON response from Gabriel's function, so we made a new flag with cleaner, formatted JSON

chrisngyn:
After talking with Benjamin, instead of using a new flag to make the JSON response pretty, we'll be using `json.tool` instead if the user chooses to return an easy to read JSON object. Will be updating documentation to reflect this alongside the new `-j` flag. 👍

chrisngyn:
@gleitz we're ready for your review!

The JSON functionality works with the `-j` flag, I updated the README accordingly to reflect this, added a note about how you can prettify the response by adding `| python -m json.tool`

I also added a short little demo .gif in the README too; if you approve of the .gif addition I can make a better one and show more of the features (this was a really short and scrappy one I made quickly)

README.rst
```diff
@@ -52,6 +52,8 @@ howdoi will answer all sorts of queries:
 
     $ howdoi create tar archive
     > tar -cf backup.tar --exclude "www/subf3" www
+    
+.. image:: howdoi.gif
```
gmelodie, line 56: Not actually a request, but I feel like this is not directly related to the json issue. (it's fine though)

howdoi/howdoi.py
```diff
@@ -332,30 +331,38 @@ def _get_links_with_cache(query):
...
+    num_answers = args['num_answers']
 
-    for answer_number in range(args['num_answers']):
-        current_position = answer_number + initial_position
-        args['pos'] = current_position
-        link = get_link_at_pos(question_links, current_position)
+    for inc_pos in range(num_answers):
+        curr_pos = inc_pos + init_pos
+        args['pos'] = curr_pos
+        link = get_link_at_pos(question_links, curr_pos)
```
gmelodie, line 351: I feel like this is not related to the json issue (more of a code styling thing). Personally I like the more explicit variable names: `answer_number`, `current_position` and `initial_position`

howdoi/howdoi.py
```diff
@@ -372,6 +379,29 @@ def _clear_cache():
...
+    @res: json object with answers and metadata
+    @args: command-line arguments (used for parsing)
+    returns: formated string of text ready to be printed
+    """
+    res = json.loads(res)
+    if "error" in res:
+        return res["error"]
+
+    splitter_length = 80
+    answer_splitter = '\n' + '=' * splitter_length + '\n\n'
```
gmelodie, line 393: In PR #280 @tayoogunbiyi added a `build_splitter` function which should be nice to use here

howdoi/howdoi.py
```diff
@@ -383,17 +413,23 @@ def howdoi(raw_query):
...
         cache.set(cache_key, res)
-
-        return res
     except (ConnectionError, SSLError):
-        return 'Failed to establish network connection\n'
+        return json.dumps({"error": "Failed to establish network connection\n"})
+    finally:
+        if args["json_output"]:
+            return res
+        else:
```
gmelodie, line 431: :scissors:  Don't need the `else` clause because the function will return if it gets into the `if`.

test_howdoi.py
```diff
@@ -26,6 +28,9 @@ def setUp(self):
         self.bad_queries = ['moe',
                             'mel']
 
+    def assertValidResponse(self, res):
+        return len(res) > 0
+
```
gmelodie, line 33: Not sure if this is related to the issue at hand. Also, what was wrong with the previous way of testing?

test_howdoi.py
```diff
@@ -43,37 +48,37 @@ def test_get_link_at_pos(self):
 
     def test_answers(self):
         for query in self.queries:
-            self.assertTrue(self.call_howdoi(query))
+            self.assertValidResponse(self.call_howdoi(query))
```
gmelodie, line 51: Again, not sure if this is related

test_howdoi.py
```diff
@@ -43,37 +48,37 @@ def test_get_link_at_pos(self):
...
         for query in self.bad_queries:
-            self.assertTrue(self.call_howdoi(query))
+            self.assertValidResponse(self.call_howdoi(query))
```
gmelodie, line 53: Again, not sure if this is related

test_howdoi.py
```diff
@@ -43,37 +48,37 @@ def test_get_link_at_pos(self):
...
 
         os.environ['HOWDOI_URL'] = 'pt.stackoverflow.com'
         for query in self.pt_queries:
-            self.assertTrue(self.call_howdoi(query))
+            self.assertValidResponse(self.call_howdoi(query))
```
gmelodie, line 57: Again, not sure if this is related

test_howdoi.py
```diff
@@ -43,37 +48,37 @@ def test_get_link_at_pos(self):
...
 
     def test_answers_bing(self):
         os.environ['HOWDOI_SEARCH_ENGINE'] = 'bing'
         for query in self.queries:
-            self.assertTrue(self.call_howdoi(query))
+            self.assertValidResponse(self.call_howdoi(query))
```
gmelodie, line 62: Again, not sure if this is related

test_howdoi.py
```diff
@@ -43,37 +48,37 @@ def test_get_link_at_pos(self):
...
         for query in self.bad_queries:
-            self.assertTrue(self.call_howdoi(query))
+            self.assertValidResponse(self.call_howdoi(query))
```
gmelodie, line 64: Again, not sure if this is related

test_howdoi.py
```diff
@@ -43,37 +48,37 @@ def test_get_link_at_pos(self):
...
 
         os.environ['HOWDOI_URL'] = 'pt.stackoverflow.com'
         for query in self.pt_queries:
-            self.assertTrue(self.call_howdoi(query))
+            self.assertValidResponse(self.call_howdoi(query))
```
gmelodie, line 68: Again, not sure if this is related

test_howdoi.py
```diff
@@ -43,37 +48,37 @@ def test_get_link_at_pos(self):
...
 
         os.environ['HOWDOI_SEARCH_ENGINE'] = ''
 
     def test_answers_duckduckgo(self):
         os.environ['HOWDOI_SEARCH_ENGINE'] = 'duckduckgo'
         for query in self.queries:
-            self.assertTrue(self.call_howdoi(query))
+            self.assertValidResponse(self.call_howdoi(query))
```
gmelodie, line 75: Again, not sure if this is related

test_howdoi.py
```diff
@@ -43,37 +48,37 @@ def test_get_link_at_pos(self):
...
         for query in self.bad_queries:
-            self.assertTrue(self.call_howdoi(query))
+            self.assertValidResponse(self.call_howdoi(query))
```
gmelodie, line 77: Again, not sure if this is related

test_howdoi.py
```diff
@@ -43,37 +48,37 @@ def test_get_link_at_pos(self):
...
 
         os.environ['HOWDOI_URL'] = 'pt.stackoverflow.com'
         for query in self.pt_queries:
-            self.assertTrue(self.call_howdoi(query))
+            self.assertValidResponse(self.call_howdoi(query))
```
gmelodie, line 81: Again, not sure if this is related

test_howdoi.py
```diff
@@ -88,7 +93,7 @@ def test_answer_links_using_all_option(self):
             self.assertNotEqual(re.match('.*http.?://.*questions/\d.*', response, re.DOTALL), None)
 
     def test_position(self):
-        query = self.queries[0]
+        query = random.choice(self.queries)
         first_answer = self.call_howdoi(query)
         not_first_answer = self.call_howdoi(query + ' -p5')
         self.assertNotEqual(first_answer, not_first_answer)
```
gmelodie, line 99: What's the difference of getting a random query and getting the first one? This could mess up testing because one query could fail and other could pass (i.e. non-deterministic outputs). Also, not sure if it's related.

test_howdoi.py
```diff
@@ -100,6 +105,13 @@ def test_all_text(self):
         self.assertNotEqual(first_answer, second_answer)
         self.assertNotEqual(re.match('.*Answer from http.?://.*', second_answer, re.DOTALL), None)
 
+    def test_json_output(self):
+        query = self.queries[0]
```
gmelodie, line 109: Inconsistent with last comment. We should either go for all `random` or all `query[0]`, although I am of the opinion that we should not use `random.choice` to test.

howdoi/howdoi.py
```diff
@@ -410,6 +446,8 @@ def get_parser():
     parser.add_argument('-n', '--num-answers', help='number of answers to return', default=1, type=int)
     parser.add_argument('-C', '--clear-cache', help='clear the cache',
                         action='store_true')
+    parser.add_argument('-j', '--json-output', help='return answers in raw json format',
```
gleitz, line 449: I like `-j` but I was looking around at some other repos and it seems like `-o json` is also somewhat common. Kubernetes uses it for `kubectl`, as done this [feature request](https://github.com/helm/helm/issues/1534). However I don't think we would ever support `-o yaml` so let's stick with `-j` for now.

test_howdoi.py
gleitz: We noticed that when we added the JSON response, all the tests started failing.

test_howdoi.py
gleitz: Technically I think `assertValidResponse` is better than `assertTrue`, and this leaves the door open to improve `assertValidResponse`. The `assertTrue` was honestly me being lazy when I wrote the tests.

test_howdoi.py
gleitz: But let's make it `self.assertTrue(len(res) > 0)` so we can use the unitest's version of assert

test_howdoi.py
gleitz: yeah please just use the first. I'm not sure if every query is going to have 5 answers but I know the first one does.

test_howdoi.py
gleitz: Let's use the first `[0]`

README.rst
chrisngyn: Whoops -- yeah you're definitely right about that, my bad 😅

howdoi/howdoi.py
YKo20010: Looks like PR #280 hasn't been merged to master yet, so we're currently waiting to make this change.

howdoi/howdoi.py
chrisngyn: Good catch -- thanks for pointing this out!

README.rst
```diff
@@ -90,6 +92,7 @@ Usage
       -a, --all             display the full text of the answer
       -l, --link            display only the answer link
       -c, --color           enable colorized output
+      -j, --json-output     return answers in raw json format, append '| python -m json.tool' at end of query to prettify
```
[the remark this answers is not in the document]
chrisngyn: I actually didn't know about using `json.tool` to pretty print JSON until you told me about it (I'm a tad new to Python still) so I think to make things more beginner friendly, we can add a note about the existence of `json.tool`, thoughts?

howdoi/howdoi.py
```diff
@@ -332,30 +331,38 @@ def _get_links_with_cache(query):
...
+        if not args['link'] and not args['json_output']:
+            star_headers = (num_answers > 1 or args['all'])
             answer = format_answer(link, answer, star_headers)
         answer += '\n'
-        answers.append(answer)
-    return answer_spliter.join(answers)
+        res.append({
+            'answer': answer, 
+            'link': link, 
+            'position': curr_pos
```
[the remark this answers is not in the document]
YKo20010: image
`position` isn't the index of the returned list, ie. if you set the initial position to be -p3 you'll see that the results start at position=3.

chrisngyn:
@gleitz ready for your review whenever you got the time!

- removed the .gif from the README and will make that into its own issue and also changed the help text for the JSON command
- went through all the comments and updated everything accordingly
- all tests passing!

howdoi/howdoi.py
```diff
@@ -18,6 +18,7 @@
 from cachelib import FileSystemCache, NullCache
 import requests
 import sys
+import json
```
gleitz, line 21: nit: please put this in alphabetical order under `cachelib`

howdoi/howdoi.py
gleitz: #280 was merged, so you can pull in those functions now (and also fix a few merge conflicts)

howdoi/howdoi.py
```diff
@@ -372,6 +378,29 @@ def _clear_cache():
     return cache.clear()
 
 
+def _parse_json(res, args):
+    """
+    @res: json object with answers and metadata
+    @args: command-line arguments (used for parsing)
+    returns: formatted string of text ready to be printed
+    """
```
gleitz, line 386: I like to keep the codebase pretty self explanatory without superfluous comments. While docstrings are nice, they're a bit out of place in this repo that I'm trying to keep succinct and readable. It's also a sign that we aren't naming our functions well.

I have a few suggestions:
1. Change `_get_instructions` to `_get_answers`.
Internally we already have the concept of `_get_answer`, so this will be a nice wrapper around that function to orchestrate the algorithm of fetching the question links and then the answers.

1. I like to use JSON like the bread in a sandwich. JSON can be read into a program, and JSON can be provided as output of a program, but the internal representation, at least for our uses, should be native objects within the language. As the code is now written, `_get_instructions` is returning JSON which must be `json.loads`'d and checked before it is returned to the user. Additionally, JSON is hard to typecheck and requires a lot of extra `json.dumps` all over the place. Let's have `_get_instructions` (now called `_get_answers`) return an array of answers.

1. Change the name of `_parse_json` to something like `_format_answers`. This function can take `answers` and `args` as parameters and convert either to a string for command line output or to JSON.

howdoi/howdoi.py
```diff
@@ -383,17 +412,21 @@ def howdoi(raw_query):
 
     res = cache.get(cache_key)
```
gleitz, line 413: After the updates I suggested above, we will be storing `answers` in the cache. We'll simply pull them out of the cache and then call `_format_answers`.

I wonder if we need to version the cache key as we make these larger changes to it. If someone has an existing cache key (which used to store strings) and now we're storing objects, we'll have issues. 

A function like `_get_cache_key` which takes a string and returns a cache key string would be useful here, as this function can deal with adding the version number.

test_howdoi.py
```diff
@@ -5,6 +5,7 @@
 import re
 import time
 import unittest
+import json
```
gleitz, line 8: nit: sort alphabetically

howdoi/howdoi.py
gleitz: Note: the JSON sandwich was inspired by the [Unicode sandwich](https://nedbatchelder.com/text/unipain/unipain.html), where you have various encodings that your program can accept and produce but everything inside the program is Unicode.
;)

gleitz:
Thanks all!

PR #280, opened 2020-06-16
Enables howdoi command

### LINKED ISSUE #279: howdoi use howdoi
Sometimes `howdoi` just works. Other times you need to phrase your question in a different way, or ask for more answers, or look at the entire stackoverflow answer, or get a link and check it out yourself. 

It might be worth augmenting the README to explain some howdoi best practices. 

My usual practice is:
- `howdoi foo`
- `howdoi foo -a` (to read the entire answer)
- `howdoi foo -n3` (to get a couple more answers in case the first wasn't the best)
- `howdoi foo -l | xargs open` (to open the stack overflow page and read it)

Perhaps a simple guide would be helpful for new users.

tayoogunbiyi:
This pull request allows users to type the command `howdoi use howdoi` and see a useful message containing commonly used howdoi commands. 

Fixes #279


README.rst
```diff
@@ -72,6 +72,20 @@ or
...
+::
+
+    howdoi use howdoi
+
+or
+
+::
+
+    howdoi howdoi
+
```
gmelodie, line 88: I feel like we should also add the output of the command here.

howdoi/howdoi.py
```diff
@@ -96,6 +96,8 @@ def _print_dbg(x): print("[DEBUG] " + x)  # noqa: E302
 CACHE_DIR = appdirs.user_cache_dir('howdoi')
 CACHE_ENTRY_MAX = 128
 
+SUPPORTED_HELP_QUERIES = ['use howdoi','howdoi']
```
gmelodie, line 99: If I call the `howdoi` function passing only "howdoi" or "use howdoi" it works, but if I pass "howdoi howdoi" or "howdoi use howdoi" it doesn't (it fetches stack overflow like any other query). I'd say it's better to include these cases as well.

howdoi/howdoi.py
```diff
@@ -381,6 +404,9 @@ def howdoi(raw_query):
     args['query'] = ' '.join(args['query']).replace('?', '')
     cache_key = str(args)
 
+    if _is_help_query(args['query']):
+        return _get_help_instructions() + '\n'
```
gmelodie, line 408: Maybe add the newline within the `_get_help_instructions` function

tayoogunbiyi:
@gmelodie thanks. Would look into these comments and address them

howdoi/howdoi.py
tayoogunbiyi: > If I call the `howdoi` function passing only "howdoi" or "use howdoi" it works, but if I pass "howdoi howdoi" or "howdoi use howdoi" it doesn't (it fetches stack overflow like any other query). I'd say it's better to include these cases as well.

Hm. Makes sense. I’d just extend the queries supported to include these

README.rst
```diff
@@ -72,6 +72,20 @@ or
 
     python setup.py install
 
+
+New to howdoi ? 
+----------------
+
+::
+
+    howdoi use howdoi
```
gleitz, line 81: maybe just howdoi howdoi. we can support all the different variants but this one is the shortest, and also the funniest.

howdoi/howdoi.py
```diff
@@ -341,9 +346,8 @@ def _get_instructions(args):
 
     answers = []
     initial_position = args['pos']
-    spliter_length = 80
-    answer_spliter = '\n' + '=' * spliter_length + '\n\n'
-
+    answer_spliter = build_splitter('=',80)
```
gleitz, line 349: Please run `flake8` on this file as there are a couple errors. Needs a space after the comma, and also two lines before each function.

howdoi/howdoi.py
gleitz: `howdoi run howdoi`
`howdoi do howdoi`

howdoi/howdoi.py
```diff
@@ -372,6 +376,25 @@ def _clear_cache():
     return cache.clear()
 
 
+def _is_help_query(query: str):
+    return any(
+            [   query.lower() == help_query for help_query in SUPPORTED_HELP_QUERIES]
```
gleitz, line 381: This could probably be a one-liner. I think it's easy enough to read.

howdoi/howdoi.py
```diff
@@ -372,6 +376,25 @@ def _clear_cache():
...
+            )
+
+
+def _get_help_instructions():
+    instruction_splitter = build_splitter(' ',60)
+    answer = [
+        'Here are a few popular howdoi commands ',
+        '>>> howdoi foo (default query)',
```
gleitz, line 389: rather than `foo`, let's use a real query so folks can follow along at home :) Ideally something that has multiple good answers so -n3 will work well.

howdoi/howdoi.py
```diff
@@ -372,6 +376,25 @@ def _clear_cache():
...
+        '>>> howdoi foo -a (read entire answer)',
+        '>>> howdoi foo -n [number] (retrieve n number of answers)',
+        '>>> howdoi foo -l (display only a link to where the answer is gotten from',
```
gleitz, line 392: Remove "gotten"

test_howdoi.py
```diff
@@ -166,7 +167,24 @@ def test_get_questions(self):
...
         self.assertSequenceEqual(actual_output, expected_output)
-
+    
+    def test_help_queries(self):
+        help_queries =  self.help_queries
+
+        for query in help_queries:
+            output = self.call_howdoi(query)
+            self.assertTrue(output)
+            assert 'retrieve n number of answers' in output
```
gleitz, line 177: I think `self.assertTrue` might be better than simply `assert`, because the former is part of the `unittest` framework.

test_howdoi.py
```diff
@@ -166,7 +167,24 @@ def test_get_questions(self):
...
+            assert 'Specify the search engine you want to use e.g google,bing,duckduckgo' in output
+    def test_help_queries_are_properly_validated(self):
```
gleitz, line 179: Be sure to also run flake8 over this file

howdoi/howdoi.py
tayoogunbiyi: There are just so many possible cases in terms of input commands intended to be query commands. For now, the commands I to the `supported_help_queries` are 
`['use howdoi','howdoi','run howdoi','do howdoi','howdoi howdoi','howdoi use howdoi']`

tayoogunbiyi:
I applied the changes as requested with a few flake8 compliance changes also @gleitz @gmelodie

howdoi/howdoi.py
```diff
@@ -49,6 +49,8 @@ def u(x):
 
 # rudimentary standardized 3-level log output
 def _print_err(x): print("[ERROR] " + x)
+
+
```
gmelodie, line 53: :bike: I guess you could trim one of the two newlines

main.py
```diff
@@ -0,0 +1,44 @@
...
+        '>>> howdoi {} -n [number] (retrieve n number of answers)',
+        '>>> howdoi {} -l (display only a link to where the answer is gotten from',
+        '>>> howdoi {} -c (Add colors to the output)',
+        '>>> howdoi {} -e (Specify the search engine you want to use e.g google,bing,duckduckgo)'
+        ]
+    
+    instructions = map(lambda s: s.format(query),instructions)
+
+    return instruction_splitter.join(instructions)
+
```
gmelodie, line 42: :bike: Same as above, I guess you can trim one line here :scissors:

howdoi/howdoi.py
```diff
@@ -397,17 +430,20 @@ def howdoi(raw_query):
 
 
 def get_parser():
-    parser = argparse.ArgumentParser(description='instant coding answers via the command line')
+    parser = argparse.ArgumentParser(
+        description='instant coding answers via the command line')
```
gleitz, line 434: There's no need to break this line. We are using 119 length as the max https://github.com/gleitz/howdoi/blob/master/.pylintrc#L321

howdoi/howdoi.py
```diff
@@ -397,17 +430,20 @@ def howdoi(raw_query):
...
     parser.add_argument('query', metavar='QUERY', type=str, nargs='*',
                         help='the question to answer')
-    parser.add_argument('-p', '--pos', help='select answer in specified position (default: 1)', default=1, type=int)
```
gleitz, line 403: Again, no need to break this line.

howdoi/howdoi.py
```diff
@@ -397,17 +430,20 @@ def howdoi(raw_query):
...
+        '-p', '--pos', help='select answer in specified position (default: 1)', default=1, type=int)
     parser.add_argument('-a', '--all', help='display the full text of the answer',
                         action='store_true')
     parser.add_argument('-l', '--link', help='display only the answer link',
                         action='store_true')
     parser.add_argument('-c', '--color', help='enable colorized output',
                         action='store_true')
-    parser.add_argument('-n', '--num-answers', help='number of answers to return', default=1, type=int)
+    parser.add_argument('-n', '--num-answers',
+                        help='number of answers to return', default=1, type=int)
```
gleitz, line 446: Don't need to break this either

howdoi/howdoi.py
```diff
@@ -441,7 +477,8 @@ def command_line_runner():
         args['color'] = True
 
     if not args['search_engine'] in SUPPORTED_SEARCH_ENGINES:
-        _print_err('Unsupported engine.\nThe supported engines are: %s' % ', '.join(SUPPORTED_SEARCH_ENGINES))
+        _print_err('Unsupported engine.\nThe supported engines are: %s' %
```
gleitz, line 480: No line break

test_howdoi.py
```diff
@@ -80,12 +81,14 @@ def test_answers_duckduckgo(self):
     def test_answer_links_using_l_option(self):
         for query in self.queries:
             response = self.call_howdoi(query + ' -l')
-            self.assertNotEqual(re.match('http.?://.*questions/\d.*', response, re.DOTALL), None)
+            self.assertNotEqual(
```
gleitz, line 84: No break

test_howdoi.py
```diff
@@ -80,12 +81,14 @@ def test_answers_duckduckgo(self):
...
+                re.match('http.?://.*questions/\d.*', response, re.DOTALL), None)
 
     def test_answer_links_using_all_option(self):
         for query in self.queries:
             response = self.call_howdoi(query + ' -a')
-            self.assertNotEqual(re.match('.*http.?://.*questions/\d.*', response, re.DOTALL), None)
+            self.assertNotEqual(
```
gleitz, line 90: No break

test_howdoi.py
```diff
@@ -98,7 +101,8 @@ def test_all_text(self):
         first_answer = self.call_howdoi(query)
         second_answer = self.call_howdoi(query + ' -a')
         self.assertNotEqual(first_answer, second_answer)
-        self.assertNotEqual(re.match('.*Answer from http.?://.*', second_answer, re.DOTALL), None)
+        self.assertNotEqual(
```
gleitz, line 104: No break

main.py
```diff
@@ -0,0 +1,44 @@
+# import os,time
```
gleitz, line 1: What is the purpose of this file?

tayoogunbiyi: this file shouldn't be here. my bad.

gleitz:
Looking great. Excellent work @tayoogunbiyi!

tayoogunbiyi:
> Looking great. Excellent work @tayoogunbiyi!

thanks!

PR #281, opened 2020-06-16
VS Code extensions version 1.0

diarreola:
Completed the first versions of the VS code extension.


diarreola:
@gleitz Can we publish the first version of the extension once you give us the ok?

gleitz:
Thanks. Would you please ignore the node_modules directory from git so this is a little easier to review?

diarreola:
of course!

PR #285, opened 2020-06-18
Mock API responses while testing

### LINKED ISSUE #277: Tests should use mocked responses
When we run tests, we make API calls to various search engines and websites. It would be better to mock these responses so we don't hit rate limits during testing. Perhaps with a flag to the test runner we can enable live calls, but this can be disabled by default.

ovezovs:
Fixes #277 
**Summary:** Eyitayo and I have implemented mock responses to API calls for the testing purposes. This fix should avoid unnecessary calls to the network and prevent us from reaching the rate limit.
Our approach is to disable the current cache at the start of testing and use requests_cache to store the local copy of the cache in a sqlite database.

**Test:** To test the fix, run test_howdoi.py both connected and disconnected to the internet.


gleitz:
Looks like some additional code snuck into this PR? It has the extension and node_modules.

ovezovs:
> Looks like some additional code snuck into this PR? It has the extension and node_modules.

I see. My bad. I will fix it soon.

juped:
This approach seems promising, but committing a 13MB binary file seems wrong (and it seems like it may include unnecessary data as well to be so large). The general idea of "hook into requests and deliver something static" makes sense to me, though. It's also unclear why it's patched in to the main `howdoi.py` program.

`requests_cache` says it monkey patches (i.e., modifies the contents of the module at runtime) itself into `requests` - perhaps there's a simpler monkey patch you can perform in `test_howdoi.py` to get the effects without the overhead?

gleitz:
If you trash your cache and then run the tests, is it still 13MB?

gleitz:
I realize that I was wrong when I mentioned requests-cache. In an earlier PR someone from the community implemented a [FileSystemCache](https://github.com/gleitz/howdoi/blob/HEAD/howdoi/howdoi.py#L102), so we'd want to use that rather than requests-cache.

Also, for this specific task we want to make sure that the "test cache" is only applied when tests are running, so there may need to be some modification of `test_howdoi.py` that installs that cache. In practice this might be quite difficult because `appdirs` looks in a different location on OSX, Windows, and Linux.

In practice, there are only two web requests that are made, and both run through the function `_get_result`.
- [fetching the page from the search engine](https://github.com/gleitz/howdoi/blob/HEAD/howdoi/howdoi.py#L218)
- [fetching the page from stackoverflow](https://github.com/gleitz/howdoi/blob/HEAD/howdoi/howdoi.py#L286)

Perhaps we can do two things:
- use a mock _get_result when tests are running
- build a function that creates a map of url -> html for every page requested during the test run. this can be pickled or gzipped and checked in. should be a lot less than 12MB. This function would be run every time you wanted to add a new test to recreate the cache.

tayoogunbiyi:
> I realize that I was wrong when I mentioned requests-cache. In an earlier PR someone from the community implemented a [FileSystemCache](https://github.com/gleitz/howdoi/blob/HEAD/howdoi/howdoi.py#L102), so we'd want to use that rather than requests-cache.
> 
> Also, for this specific task we want to make sure that the "test cache" is only applied when tests are running, so there may need to be some modification of `test_howdoi.py` that installs that cache. In practice this might be quite difficult because `appdirs` looks in a different location on OSX, Windows, and Linux.
> 
> In practice, there are only two web requests that are made, and both run through the function `_get_result`:
> 
> * [fetching the page from the search engine](https://github.com/gleitz/howdoi/blob/HEAD/howdoi/howdoi.py#L218)
> * [fetching the page from stackoverflow](https://github.com/gleitz/howdoi/blob/HEAD/howdoi/howdoi.py#L286)
> 
> Perhaps we can do two things:
> 
> * use a mock _get_result when tests are running
> * build a function that creates a map of url -> html for every page requested during the test run. this can be pickled or gzipped and checked in. should be a lot less than 12MB. This function would be run every time you wanted to add a new test to recreate the cache.

The approach of building a function that maps from URL -> HTML seems like the way to go and is something we've previously considered.
I'd look into that with @ovezovs throughout the day.

ovezovs:
This PR is ready for review

juped:
Nice monkey-patching of the method. I think you got rate limited and some errors snuck into your cache (unless those were intentional errors).

ovezovs:
> Nice monkey-patching of the method. I think you got rate limited and some errors snuck into your cache (unless those were intentional errors).

Just checked those error pages and I believe they are needed.

howdoi/howdoi.py
```diff
@@ -139,6 +141,11 @@ def get_proxies():
     return filtered_proxies
 
 
+def format_url_to_filename(url, file_ext='html'):
```
gleitz, line 144: Let's add a leading `_` to show that this is an internal function.

test_howdoi.py
```diff
@@ -14,7 +15,28 @@ class HowdoiTestCase(unittest.TestCase):
     def call_howdoi(self, query):
         return howdoi.howdoi(query)
 
+    def mocked_get_result(self, url):
```
gleitz, line 18: Let's call this `_get_result_mock`

ovezovs: okay. working on it

gleitz:
Thanks all, this is great!

Issue #275, opened 2020-06-12
Plugin integration


gleitz:
At the core of howdoi is two steps:
- Searcher that knows how to craft a search engine query
- Extractor that knows how to extract the relevant section of the page that is found.

If we want to be able to support plugins for new sites (wikihow) and search categories (recipes), we should have an extensible plugin tool for writing and enabling these plugins.

PR #287, opened 2020-06-23
Implementing research option in howdoi to search articles from arXiv API

csr:
Not meant to be merged, just for feedback.

### DIFF
--- a/howdoi/arxiv.py
+++ b/howdoi/arxiv.py
@@ -2,7 +2,9 @@ import xml.etree.ElementTree as ET
 from urllib.parse import urlencode
 import requests
 
-
+# The goal of the API is to allow application developers access to all of the arXiv data,
+# search and linking facilities with an easy-to-use programmatic interface.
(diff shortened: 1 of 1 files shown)


howdoi/arxiv.py
```diff
@@ -0,0 +1,56 @@
+import xml.etree.ElementTree as ET
+from urllib.parse import urlencode
+import requests
+
+
+ROOT_URL = 'http://export.arxiv.org/api/'
```
ovezovs, line 6: Looks nice and clean. My only recommendation would be adding some API specific comments that will help a new contributor understand what the API is about.

csr: Thanks! Any more changes (even minor nitpicks), let me know 💯

csr:
Closing as we're looking to make the most of the howdoi built-in searcher (instead of querying the arXiv API, Google may give better results). Related to #275

PR #289, opened 2020-06-24
Design plugin feature

tayoogunbiyi:


csr:
Hi @juped & everyone, Eyitayo, Mwiza, and I have been working on a solution to decompose the howdoi codebase into `howdoi.py`, `BasePlugin` (base class for all plugins) and `StackOverflowPlugin` (a subclass of `BasePlugin` which is used as the default plugin for the user query if no plugins are specified).

We'd appreciate feedback on:
* The general code architecture
* Coding style and general improvements
* Suggestions on how users will be able to install third-party plugins (we thought about having a system that fetches plugins from your GitHub account). In general though, users will be able to manually add plugin files in the plugins folder
* Where the `cache` needs to be stored. Should it be a global variable that all plugins can access?
* Everything else that comes to your mind :)

howdoi/plugins/base.py
```diff
@@ -0,0 +1,166 @@
...
+                return True
+        return False
+
+    def _add_links_to_text(self, element):
+        hyperlinks = element.find('a')
+
+        for hyperlink in hyperlinks:
+            pquery_object = pq(hyperlink)
+            href = hyperlink.attrib['href']
+            copy = pquery_object.text()
```
teymour-aldridge, line 99:
```suggestion
            copy = xml_extractor.text()
```

howdoi/plugins/base.py
```diff
@@ -0,0 +1,166 @@
...
+            if (copy == href):
+                replacement = copy
+            else:
+                replacement = "[{0}]({1})".format(copy, href)
+            pquery_object.replace_with(replacement)
```
teymour-aldridge, line 104:
```suggestion
            xml_extractor.replace_with(replacement)
```

howdoi/plugins/base.py
```diff
@@ -0,0 +1,166 @@
...
+
+    def get_link_at_pos(self, links, position):
```
teymour-aldridge, line 106: This can be a static method.

howdoi/plugins/base.py
```diff
@@ -0,0 +1,166 @@
...
+        self._add_links_to_text(element)
+        try:
+            return element.text(squash_space=False)
+        except TypeError:
+            return element.text()
+
+    def _get_search_url(self, search_engine):
+        return SEARCH_URLS.get(search_engine, SEARCH_URLS['google'])
+
+    def _extract_links_from_bing(self, html):
```
teymour-aldridge, line 126: This can also be a static method.
```suggestion
    @staticmethod
    def _extract_links_from_bing(html):
```

howdoi/plugins/base.py
```diff
@@ -0,0 +1,166 @@
...
+
+    def get_text(self, element):
+        ''' return inner text in pyquery element '''
+        self._add_links_to_text(element)
+        try:
+            return element.text(squash_space=False)
+        except TypeError:
+            return element.text()
+
+    def _get_search_url(self, search_engine):
```
teymour-aldridge, line 123:
```suggestion
    @staticmethod
    def _get_search_url(search_engine):
```

howdoi/plugins/base.py
```diff
@@ -0,0 +1,166 @@
...
+
+    def _extract_links_from_bing(self, html):
+        html.remove_namespaces()
+        return [a.attrib['href'] for a in html('.b_algo')('h2')('a')]
+
+    def _extract_links_from_google(self, html):
+        return [a.attrib['href'] for a in html('.l')] or \
+            [a.attrib['href'] for a in html('.r')('a')]
+
+    def _extract_links_from_duckduckgo(self, html):
```
teymour-aldridge, line 134: This can also be a static method
```suggestion
    @staticmethod
    def _extract_links_from_duckduckgo(html):
```

howdoi/plugins/base.py
```diff
@@ -0,0 +1,166 @@
...
+        results = []
+        for anchor in links_anchors:
+            link = anchor.attrib['href']
+            url_obj = urlparse(link)
+            parsed_url = parse_qs(url_obj.query).get('uddg', '')
+            if parsed_url:
+                results.append(parsed_url[0])
+        return results
+
+    def _extract_links(self, html, search_engine):
```
teymour-aldridge, line 146:
```suggestion
    @staticmethod
    def _extract_links(html, search_engine):
```

howdoi/plugins/base.py
```diff
@@ -0,0 +1,166 @@
...
+        for key, value in proxies.items():
+            if key.startswith('http'):
+                if not value.startswith('http'):
+                    filtered_proxies[key] = 'http://%s' % value
+                else:
+                    filtered_proxies[key] = value
+        return filtered_proxies
+
+    def extract(self):
+        print("Hello extract")
```
teymour-aldridge, line 165: I'm not sure why this is here.
If it's intended for it to be overriden in child classes, I'd suggest either raising a `NotImplementedError` or just leave it as it is (without the print statement).

howdoi/plugins/stackoverflow.py
```diff
@@ -0,0 +1,264 @@
...
+USER_AGENTS = ('Mozilla/5.0 (Macintosh; Intel Mac OS X 10.7; rv:11.0) Gecko/20100101 Firefox/11.0',
+               'Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:22.0) Gecko/20100 101 Firefox/22.0',
+               'Mozilla/5.0 (Windows NT 6.1; rv:11.0) Gecko/20100101 Firefox/11.0',
+               ('Mozilla/5.0 (Macintosh; Intel Mac OS X 10_7_4) AppleWebKit/536.5 (KHTML, like Gecko) '
+                'Chrome/19.0.1084.46 Safari/536.5'),
+               ('Mozilla/5.0 (Windows; Windows NT 6.1) AppleWebKit/536.5 (KHTML, like Gecko) Chrome/19.0.1084.46'
+                'Safari/536.5'), )
+
+
+def _random_int(width):
```
teymour-aldridge, line 32: Assuming that this just generates a random number, would it be easier to use `random.randint`?

howdoi/plugins/base.py
```diff
@@ -0,0 +1,166 @@
...
+URL = os.getenv('HOWDOI_URL') or 'stackoverflow.com'
+
+SEARCH_URLS = {
+    'bing': SCHEME + 'www.bing.com/search?q=site:{0}%20{1}&hl=en',
+    'google': SCHEME + 'www.google.com/search?q=site:{0}%20{1}&hl=en',
+    'duckduckgo': SCHEME + 'duckduckgo.com/?q=site:{0}%20{1}&t=hj&ia=web'
+}
+
+
+class BasePlugin():
```
teymour-aldridge, line 81: Architecture thoughts:

```python
import enum.Enum as Enum

class Source(Enum):
   CLI = 0
   DISCORD_MESSAGE = 1

class ServiceConnectionError(Exception):
  pass

class Plugin:
    name = "UnimplementedPlugin"
    def __init__():
      pass
    def resolve(self):
      pass
    def extract(self, source, data): # raises an exception if this fails
      pass

class PluginStack(list):
   def __init__(self, plugins):
     super(PluginStack, self).__init__(plugins)
   def handle_input(self, source, data):
     tries = [ ]
     for plugin in self:
       try:
         plugin.extract(source, data)
         plugin.resolve()
       except <any of the relevant exceptions>:
         tries.append("howdoi could not get a response from {}".format(plugin.name))
```

Further possible areas for improvement:
 * adding `async` support.
 * "real-time" error handling – i.e. if the first plugin fails, howdoi immediately sends a response
 * optional "supports" attribute for `Plugin`s which is a list containing all the sources they support.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "python.pythonPath": "/Users/cesaredecal/workspace/Environments/howdoi/bin/python"
```
teymour-aldridge, line 2: You probably shouldn't check this in and instead add it to your .gitignore.

csr: Thank you! I forgot

teymour-aldridge: Np.

howdoi/plugins/base.py
csr: I'm trying to understand `PluginStack`. Is it the manager of the plugins installed in howdoi?

teymour-aldridge: Yes, that's the idea. Whenver it receives a request it tries each plugin in turn. If a plugin fails, it moves on to the next plugin, otherwise it returns the result of that plugin. The code above isn't 100% complete.

csr: Thank you. I think the user experience is having the user specify the plugin they want to use with a flag along with their query (at least for now, Hajer is working on a model that automatically detects which plugin is the most appropriate). How would things change then?

teymour-aldridge: I think that there has to be a list of installed plugins and a way of resolving/calling them somehow. A class to handle this could still be useful, but it would be a bit different to the `PluginStack`.

howdoi/plugins/base.py
csr: This feedback is great, we'd like to keep this PR a "refactor" and save these valuable renames for a later PR so that we don't change the previous code too much

PR #290, opened 2020-06-26
Imlementing JSON plugins

csr:
This patch enables support for JSON plugins 🎉  Two sample plugins have been included to demonstrate how it works:
* arXiv: to search scholarly articles (returns abstract of article)
* Wikipedia: to search Wikipedia (returns first paragraph of article)

This is a sample plugin JSON file that enabled `howdoi` to work with Wikipedia: 

```
{
    "url": "https://en.wikipedia.org/wiki",
    "content_selector": "p:first"
}
```

`url` filters the search engine results, whereas `content_selector` selects the content that needs to be parsed from any HTML page. Plugins are stored inside the plugins folder. `-n NUM_ANSWERS` is supported. 

A few things that we think could be improved:
* How plugins are stored and used in the codebase. Right now we call `parse_plugin` each time we need it, but can we find a way to only parse it (and handle all the possible exceptions) _once_?
* Enable the plugin to specify which websites from the search engine search should not be considered when extracting links.
* Enable the plugin to support the `-a` or `--all` flags (perhaps by defining additional content selectors to select the _entire_ text)
* Make it easier to install plugins (perhaps download them from a GitHub repository or url that you specify)

@tayoogunbiyi, @rkbeatss, @csr


rkbeatss:
I have added some tests today but I definitely still need to change it and refine it more. Here's a list of what we still need to do for tomorrow:

- [ ]  Discuss and fix caching bug

- [ ] Refine tests (Will be working on it!)

- [ ]  Only add answertab=votes (search in code) if args['url'] is stackoverflow 

- [X] Documentation via read me

howdoi/howdoi.py
```diff
@@ -453,6 +453,11 @@ def howdoi(raw_query):
         parser = get_parser()
         args = vars(parser.parse_args(raw_query.split(' ')))
 
+    if args.get('plugin'):
```
csr, line 456: Maybe we might want to keep the plugin parsing _after_ the optional cache clearance code, because if the user decided to clean the cache then the program would clear the cache and stop (we're not interested in parsing the plugin).

test_howdoi.py
```diff
@@ -49,6 +50,7 @@ def setUp(self):
                            'hello world em c']
         self.bad_queries = ['moe',
                             'mel']
+        self.plugin_queries = ['robotics', 'computing']
```
tayoogunbiyi, line 53: Nice test. however, I think the name for plugin_queries can be changed? I'm kinda thinking it's not exactly a 'query'.

howdoi/plugins/arxiv.json
```diff
@@ -0,0 +1,4 @@
+{
+    "url": "https://arxiv.org/abs",
+    "selector": ".abstract.mathjax"
```
juped, line 3: You probably want to select "abstract" here.

howdoi/howdoi.py
```diff
@@ -318,28 +318,57 @@ def _get_answer(args, links):
...
+        return False
+
+    cache_key = link
+    page = cache.get(link)
+    if not page:
+        if args['url'] == STACKOVERFLOW_URL:
+            extendend_link = link + '?answertab=votes'
+        else:
+            extendend_link = link
+        page = _get_result(extendend_link)
```
juped, line 336: do you mean "extended"?

howdoi/howdoi.py
```diff
@@ -318,28 +318,57 @@ def _get_answer(args, links):
         text = '\n'.join(texts)
     else:
         text = _format_output(get_text(instructions.eq(0)), args)
+    return text
```
juped, line 321: Can Stack Overflow be made a plugin (hardcoded dict, in this case)? If not, then why not? (You might not want to, but on the other hand it might point towards later enhancements to the plugin system.)

howdoi/howdoi.py
```diff
@@ -476,8 +513,24 @@ def get_parser():
...
     return parser
 
+def parse_plugin_from_filepath(path):
+    try:
+        with open(path) as json_file:
+            plugin = {}
+            json_file_content = json.load(json_file)
+            plugin['url'] = json_file_content['url']
+            plugin['selector'] = json_file_content['selector']
+            return plugin
```
juped, line 526: Returning a dict is good. We can consider the dict to be the real "plugin", and the file to be an expression of it. Maybe other filetypes are possible.

test_howdoi.py
```diff
@@ -106,7 +110,57 @@ def test_answers_duckduckgo(self):
             self.assertValidResponse(self.call_howdoi(query))
 
         os.environ['HOWDOI_SEARCH_ENGINE'] = ''
-
+    
+    def create_plugin_file(self, filename, content):
+        path = os.path.join(howdoi.PLUGIN_DIR, filename)
+        with open(path, 'w+') as f:
+            json.dump(content, f)
```
juped, line 117: Can we expose this to the user to let them generate plugins after just inspecting an element?

howdoi/plugins/wikipedia.json
```diff
@@ -0,0 +1,4 @@
+{
+    "url": "https://en.wikipedia.org/wiki",
+    "selector": "p:first"
```
juped, line 3: Does this work when Wikipedia has its donation begging banner up?

test_howdoi.py
csr: What do you mean with this?

csr:
Closing for now, anyone else feel free to keep working on it.

PR #292, opened 2020-06-29
MKdocs setup

IamCathal:
I added the basic MKDocs setup files along with some examples of the features you can use like tooltips, code blocks etc. This isn't meant to be a finished product it's mainly a preview as to what MKDocs looks like and how to use the various features and extensions that it has.

The commands you need to use it are
`pip install mkdocs-material markdown-include`
and
`python -m mkdocs serve` - Start the live-reloading docs server.
`python -m mkdocs build` - Build the documentation site.
`python -m mkdocs help` - Print this help message

At the moment there is no live buiilding and serving of the docs on commits as there would be with the github actions setup for it. I took some inspiration from [Typer's](https://github.com/tiangolo/typer) use of MKDocs so if you'd like to get that setup it'd be a good place to get started from in terms of seeing how it all works.

[Here](https://mlh-fellowship.github.io/howdoi/)'s the live version of the docs.

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -1,0 +1,207 @@
+## General guidelines
+
+Be sure to go through these items before creating a new issue:
+
+1. Check the [existing issues](https://github.com/gleitz/howdoi/issues) to see if anyone is already working or have already worken on your issue or a similar one.
+
+2. If there are no current or past issues similar to yours, be sure to give a a **complete description** when creating it.
+
+3. Wait for feedback on the issue before starting to work.
+
+!!! tip
+    Include instructions on how to reproduce the bug you found or specific use cases of a requested feature.
+
+!!! Note
+    Follow Github's [guide to collaborating efficiently](https://lab.github.com/githubtraining/introduction-to-github).
+
+
+
+## Setting up development environment
+
+Clone the git repository
+```bash
+$ git clone https://github.com/gleitz/howdoi.git
+```
+
+Setup and activate a virtual environment
+```bash
+$ python3 -m venv .venv
+$ source .venv/bin/activate
+```
+
+Install packages
+```bash
+$ pip install -r requirements.txt
+```
+
+## Running howdoi
+
+Run on the command-line
+```bash
+python -m howdoi.howdoi QUERY
+```
+
+!!! note
+    If you try running `python howdoi/howdoi.py` (without `-m`) you might get `ValueError: Attempted relative import in non-package`.
+
+If you want to use howdoi from within a python script, just pass your query to `howdoi.howdoi()`
+
+```python
+from howdoi import howdoi
+
+query = "for loop python"
+output = howdoi.howdoi(query)
+```
+
+Or parse it yourself and passed the arguments to `howdoi.howdoi()`
+```python
+from howdoi import howdoi
+
+query = "for loop python"
+parser = howdoi.get_parser()
+args = vars(parser.parse_args(query.split(' ')))
+
+output = howdoi.howdoi(args)
+```
+
+!!! attention
+    Parsing queries yourself is the older way to pass in queries and may be deprecated in the future. Prefer the first example.
+
+
+## Submitting Pull Requests
+Before PRs are accepted they must pass all [Travis tests](https://travis-ci.org/gleitz/howdoi) and not have any `flake8` or `pylint` warnings or errors.
+
+#### Testing
+Howdoi uses python's [`unittest`](https://docs.python.org/3/library/unittest.html) library for unit testing. Run the unit tests locally
+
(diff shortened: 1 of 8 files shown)


docs/contributing.md
```diff
@@ -0,0 +1,10 @@
+### How can I help contribute to howdoi?
+
+If you would like to add or extend existing functionality, fix a bug or anything inbetween please follow the standard procedure for most open source projects. i.e:
+
+1. First check the issues to see if anyone is working on your issue before starting work yourself. Someone else might be working on it already or a maintainer might've closed the issue as they don't feel it would be beneficial to the project.
+2. Submit the issue on the repo and begin your work on your fork of the repo.
+3. Submit your PR once completed and wait for feedback.
+
+!!! Note
```
ovezovs, line 9: For this, I would suggest adding a colon and bolding `**Note:**`. What do you think?

docs/contributing.md
```diff
@@ -0,0 +1,10 @@
+### How can I help contribute to howdoi?
+
+If you would like to add or extend existing functionality, fix a bug or anything inbetween please follow the standard procedure for most open source projects. i.e:
+
+1. First check the issues to see if anyone is working on your issue before starting work yourself. Someone else might be working on it already or a maintainer might've closed the issue as they don't feel it would be beneficial to the project.
```
ovezovs, line 5: Just a nit-picky thing but something to consider: I think `First` can be removed since we are using numbers

docs/installation.md
```diff
@@ -0,0 +1,22 @@
...
+pip install howdoi
+``
+
+or
+
+``
+pip install git+https://github.com/gleitz/howdoi.git#egg=howdoi
+``
+
+!!! Note
```
ovezovs, line 15: I would suggest a similar change to Note as I said above just to keep the consistency

gmelodie:
@ovezovs Thanks for the review! I'll take a look at those. I think it's best to leave comments for when the PR is completely ready though :)

ovezovs:
> @ovezovs Thanks for the review! I'll take a look at those. I think it's best to leave comments for when the PR is completely ready though :)

 @gmelodie, I didn't realize that it wasn't ready, my bad. I may have misunderstood Cathal's message in the howdoi channel.

gleitz:
Cool, looks neat. Let me know when you have something basic hosted so I can take a look at it visually.

gmelodie:
Oh we completely forgot to link that @gleitz. [Here's the live version](https://mlh-fellowship.github.io/howdoi/).

gmelodie:
@juped if you could have a look that'd be great!

docs/contributing.md
```diff
@@ -0,0 +1,207 @@
...
+
+#### Project layout
+
+    mkdocs.yml    # The configuration file.
+    docs/
+        index.md  # The documentation homepage.
+        ...       # Other markdown pages, images and other files.
+
+
+#### Here are some example alerts you can use
```
gleitz, line 130: I think we can remove this section?

docs/troubleshooting.md
```diff
@@ -0,0 +1,18 @@
+You might get the following error when installing with Homebrew:
```
gleitz, line 1: We end up with some code duplication between each of these individual sections and the main README.rst.

I think the reason I used an `rst` was because pypi used to not support markdown very well. I'm pretty sure it has good support now. We can convert the README.rst to markdown, but I wonder if we can ever get rid of the documentation duplication?

docs/contributing.md
teymour-aldridge: I think this is a custom syntax extension to Markdown. https://squidfunk.github.io/mkdocs-material/reference/admonitions/

docs/troubleshooting.md
gmelodie: We can make the new `README.md` point to the documentation, much like [this README](https://github.com/pallets/werkzeug)

docs/user-guide.md
```diff
@@ -0,0 +1,36 @@
...
+$ howdoi howdoi
+```
+
+
+Print the help manual
+```bash
+$ howdoi # "howdoi -h" also prints help
+
+
+usage: howdoi.py [-h] [-p POS] [-a] [-l] [-c] [-n NUM_ANSWERS] [-C]
```
gleitz, line 15: Please update this with the latest version of the usage from the main README (there's a couple new commands).

mkdocs.yml
```diff
@@ -0,0 +1,42 @@
+site_name: HowDoI
```
gleitz, line 1: Replace with `howdoi` (and do a quick grep for HowDoI elsewhere)

PR #294, opened 2020-06-30
Change test teardown methods

tayoogunbiyi:
@gleitz This replaces the teardown method with deleting the keys from the os.environ object. it's important to check if the key exists before attempting to delete because the env variable may not be set.


gleitz:
Note: this also speeds up the tests because we're removing the sleep!

Issue #297, opened 2020-07-22
howdoi errors out if keep directory has not been initialized


chrisngyn:
To use the stashing feature in `howdoi`, we rely on `keep`. To use `keep`, you need to run `keep init` first to initialize your `.keep/` directory. However, you can use `keep` without `keep init`, if you just type `keep new` the program is smart enough to initialize the folder for you. `howdoi` **skips over this intermediate step.**

PR #298, opened 2020-07-22
Fixed keep initialization

chrisngyn:
Addresses #297.

Simple check to see if a `FileNotFoundError` exception would be thrown or not. If thrown, we know the user did not initialize `.keep/` so we run `keep init` to initialize the environment for them. After we do this, we save their query.

If a `FileNotFoundError` exception is not thrown, then we're good to go.


gleitz:
Very nice thank you!

PR #299, opened 2020-07-23
Vscode Extension + Extension Plugin 

diarreola:
This contains the howdoi visual studio code extension and extension plugin enabling the json flag. This is an updated version of the previous pr.


extension/ext-plugin/package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "ext-plugin",
+  "version": "1.0.0",
+  "description": "",
+  "main": "plugin.js",
+  "scripts": {
+    "test": "cross-env TS_NODE_FILES=true mocha --exit --require ts-node/register --colors src/test/**/*.ts",
```
gleitz, line 7: So cool!

extension/ext-plugin/src/plugin.ts
```diff
@@ -0,0 +1,153 @@
+'use strict'
+import * as cp from 'child_process'
+
+const HOWDOI_PREFIX = 'howdoi'
+
+interface HowdoiObj {
+  question: string
+  answer: string[]
+  link: string[]  
+}
```
gleitz, line 10: Very nice typing

extension/ext-plugin/src/test/plugin.test.ts
```diff
@@ -0,0 +1,159 @@
...
+      // LaTex single line comment
+      assert.equal(plugin.addComment('howdoi query', commentChar3), '% howdoi query')
+      // Add coment: -- 
+      const commentChar4 = { frontComment: '--', endComment: '' }
+      assert.equal(plugin.addComment('howdoi query', commentChar4), '-- howdoi query')
+      // Add comment: //
+      const commentChar5 = { frontComment: '//', endComment: '' }
+      assert.equal(plugin.addComment('howdoi query', commentChar5), '// howdoi query')
+    })
+  })
```
gleitz, line 155: Good use of tests here. As a potential future improvement, you could perhaps generate the test string from the `commentChar` object, e.g.

`{ frontComment: '//', endComment: '' }` -> `'// howdoi query'`

And then have a loop over the `commentChar` objects.

extension/ext-plugin/tsconfig.json
```diff
@@ -0,0 +1,14 @@
...
+    "outDir": "out",                    /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', 'es2020', or 'ESNext'. */
+    "lib": ["es2018", "dom"],                             /* Specify library files to be included in the compilation. */
+    "sourceMap": true,                     /* Generates corresponding '.map' file. */
+    "rootDir": "src",
+    "strict": true,                           /* Enable all strict type-checking options. */
+    "esModuleInterop": true,                  /* Enables emit interoperability between CommonJS and ES Modules via creation of namespace objects for all imports. Implies 'allowSyntheticDefaultImports'. */
+    "skipLibCheck": true,                     /* Skip type checking of declaration files. */
+    "forceConsistentCasingInFileNames": true  /* Disallow inconsistently-cased references to the same file. */
+  }
+}
```
gleitz, line 14: In practice we can probably remove the comments from this file.

diarreola:
For the `ext-plugin` folder, I was thinking of renaming it to `code-editor-integration` to be more descriptive. What do you think? In addition, within the flow of the root readme, where would you like this development description? Currently, I have it under the `Development` section as a section named `Extension`. Within that section, I'm hoping to cover your questions and have a subsection named `Visual Studio Code howdoi Extension` to explain the two different installations (packaged file and marketplace installation).

gleitz:
Yes, I think it's a great idea to be more descriptive with the rename. If you have `code-editor-integration` and `vscode-howdoi`, maybe one can depend on the other so there isn't duplicated code? Or perhaps the code duplication comes from needing to package one for the store. I'll wait for your full README in the `Extension` section :)

diarreola:
They're not entirely duplicated. The `vscode-howdoi` folder holds the code for the vscode extension. For this, I added some code for a drop down menu and used the vscode api. And yes, the vscode extensions still needs to be packaged!
The `code-editor-integration` folder holds files that simplify the process of integrating howdoi as a code editor extension. Just finished the README and unsure if I went overkill for the `Code editor integration commands`.

gleitz:
I wonder if we can have a shared folder with common code that both the vscode and generic extension would share?

diarreola:
I can definitely create a wrapper function within the vscode-ext file to call the plug-in file so there is no duplicate code. It seems like I will need to create a copy of the `code-editor-integration` folder and place it in my vscode-ext folder to package it and add it to the marketplace. Does this seem right to you? Please let me know if I'm misunderstanding something :)

gleitz:
Ideally you would have both as separate folders, and then somehow include `code-editor-integration` from `vscode-howdoi`. I wonder if something like this would work:

```
// within vscode-howdoi/src/extension.ts
import * as plugin from '../../code-editor-integration/src/plugin'
```

It might not work because you're going up a level higher than the package.json? If that's the case, I could image that as part of an npm script you could do the copy, and then have that folder be ignored so it won't get checked in:

```
"copy-folder": "cp ../code-editor-integration .",
```

Then `vscode-howdoi` becomes a very thin wrapper around `code-editor-integration` which is doing the heavy lifting. That's nice because then for the Atom or Emacs plugin it can call out to the same `code-editor-integration`

I'm still a little unsure what exactly gets produced in the end for VS Code. According to [this article](https://code.visualstudio.com/api/working-with-extensions/publishing-extension#packaging-extensions) we'll end up with a `.vsix` file, but maybe that's just the `./out/extension.js` file plus the icon and some other things.

diarreola:
Ready for review :)

gleitz:
> Ready for review :)
> Quick Q: In the `vscode-howdoi/package.json` file, who do you want me to put for publisher name?

That can be you! (if you like)

gleitz:
Quick nit: looks like there is a conflict with `README.rst` so you'll want to do a git pull or a rebase to get that up to date with master.

extension/code-editor-integration/src/test/plugin.test.ts
```diff
@@ -0,0 +1,101 @@
...
+  // <!-- -->: HTML, PHP, Markdown, Vue single line comment
+  const commentChar4 = { frontComment: '%', endComment: '' }
+  // --: SQL, Haskell single line comment
+  const commentChar5 = { frontComment: ';', endComment: '' }
+  // %: LaTex single line comment
+  const commentChar6 = { frontComment: '/*', endComment: '*/' }
+  // ;: clojure single line comment
+  const commentChar7 = { frontComment: '<!--', endComment: '-->' }
+  /* eslint-disable prefer-const*/
+  const commentCharArr = [commentChar1, commentChar2, commentChar3, commentChar4, commentChar5, commentChar6,
```
gleitz, line 22: Very clean here. Nice work.

extension/vscode-howdoi/package.json
```diff
@@ -0,0 +1,55 @@
...
+    "commands": [
+      {
+        "command": "howdoi.extension",
+        "title": "howdoi"
+      }
+    ]
+  },
+  "scripts": {
+    "copy": "ncp ../code-editor-integration/ src/code-editor-integration/",
+    "clean": "rimraf ./src/code-editor-integration/node_modules",
```
gleitz, line 27: Cool, I didn't know about `rimraf`.

extension/vscode-howdoi/package.json
```diff
@@ -0,0 +1,55 @@
...
+  "contributes": {
+    "commands": [
+      {
+        "command": "howdoi.extension",
+        "title": "howdoi"
+      }
+    ]
+  },
+  "scripts": {
+    "copy": "ncp ../code-editor-integration/ src/code-editor-integration/",
```
gleitz, line 26: I realize this is sort of a hack, but it makes the codebase so much more clean!

extension/vscode-howdoi/package.json
diarreola: I removed `code-editor-integration/node_modules` from the `vscode` files since it was producing an error about file duplication when testing the `vscode` files.

diarreola:
I have two more questions:
1. I noticed that when running the extension, a user cannot ask `howdoi howdoi` since it's not of type json obj. I think this brings up the question if users should be able to use flags within the code editor, since this query provides examples of how to use the different flags. Which would be cool actually! But I would need to change a few things :) Currently the extension doesn't support adding flags to a query and gives weird answers (below is an example of using the flag `-a`). Would love to know what you think about this.
Screen Shot 2020-08-06 at 1 33 29 PM


2. My GitHub workflow CI is currently failing for the vscode files when running the test files. This is because `/code-editor-integration/src/test` cannot be found with the `vscode-ext` file (we're not pushing that file). I can remove the testing step for the `vscode-ext` in the workflow to remove this error. Is that ok?

diarreola:
Also, I can clean up the 300+ commits today!

gleitz:
LGTM! Ready to merge when you are

diarreola:
I have fixed the GitHub Actions workflow error, thank you for the help! I have also updated the `README` in the `code-editor-integration` file per your suggestion which includes how to use the plug-in and a description of the input parameter and the returned object for the plug-in. Let me know if there is anything else you would like me to update :) 
I know you mentioned before making a small command line wrapper for the plug-in. Would you like me to include that now or in another pr?

diarreola:
I've recently, 
1. refactored the plugin files, 
2. enabled the `-n` flag allowing users the option to receive more than 3 answers (default: 3)
3. Removed inline expressions such as, `>>>` and `...` from the answers, so code is presented cleaner in the code editor
for example, removing the expressions in this case or others like it:
Screen Shot 2020-08-11 at 8 32 39 AM

gleitz:
🎉 🎉 🎉 LOVE IT!

PR #302, opened 2020-07-30
Reduce cache size

tayoogunbiyi:
This PR looks to reduce the size of the earlier `cache_html` folder. By using the inbuilt `gzip` library to compress the HTML files into binary, we reduced the cache size from `11.1MB` to `2.3MB`. 

I also renamed the cache folder as since the content isn't HTML anymore, the folder name turns out to be misleading if I leave it as `cache_html`

### DIFF
--- a/test_howdoi.py
+++ b/test_howdoi.py
@@ -5,7 +5,6 @@ import gzip
 import json
 import os
 import re
-import time
 import unittest
 
 from cachelib import NullCache


tayoogunbiyi:
This snippet which I used to transform all the existing HTML files into .gz files wasn't included in the codebase so I decided to add it on here.

```
def transform(filename):
    orig_filepath = os.path.join("cache_html", filename)
    with open(orig_filepath, "rb") as f_in:
        new_file_name = os.path.join("cache_html_gzip", filename + ".gz")
        with gzip.open(new_file_name, "wb") as f_out:
            shutil.copyfileobj(f_in, f_out)
```

gleitz:
This is great! I wonder if it's possible to store the entire cache as one `.gz` file?

gleitz:
On second thought, I think it's actually more readable to see all the pages that get cached. So we can leave it like this :)

PR #303, opened 2020-08-07
User stats

tayoogunbiyi:


gleitz:
This is looking really good!

gleitz:
I think if we take the basic stats you're already collecting, show them in a neat way on the command line, and add a couple command line arguments for viewing and clearing the stats this can be in a good place to release a first version

Issue #310, opened 2020-09-02
Adding -a flag changes output to "not found"


mcint:
```
$ howdoi c compiler error on warn
#ifndef DEBUG
#error Only Debug builds are supported
#endif
```
```
$ howdoi c compiler error on warn -a
Sorry, couldn't find any help with that topic
```
```
$ howdoi c compiler error on warn -l
https://stackoverflow.com/questions/2221517/how-do-i-generate-an-error-or-warning-in-the-c-preprocessor
```
And the non-sensical, or not useful.
```
$ howdoi c compiler error on warn -l -a
Sorry, couldn't find any help with that topic

PR #311, opened 2020-09-02
fit new stackoverflow css class

WindSoilder:
Fixes: #310 
This issue is because `stackoverflow` have change the answer node relative `class`.  And we can't find `.post-text` anymore.

### DIFF
--- a/howdoi/howdoi.py
+++ b/howdoi/howdoi.py
@@ -313,18 +313,24 @@ def _get_answer(args, links):
         cache.set(cache_key, page)
 
     html = pq(page)
-    print(page)
 
     first_answer = html('.answer').eq(0)
 
     instructions = first_answer.find('pre') or first_answer.find('code')
     args['tags'] = [t.text for t in html('.post-tag')]
 
+    # make decision on answer body class.
+    if first_answer.find(".js-post-body"):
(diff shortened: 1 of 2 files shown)


WindSoilder:
The UT runs into failed... I need to take more time to figure out what's going on here.. sorry for the disturb.

WindSoilder:
Sometimes stackoverflow will return answer body with class `js-post-body` rather than `post-tag`.   
图片
So this PR will try to detect the class of answer body node first.

For the `js-post-body` case, I have tried to use it in another repo (here is the relative [PR](https://github.com/WindSoilder/hors/pull/88/files)), and it works fine for me :-)

gleitz:
Thank you for this!

gleitz:
I'm realizing that this change requires a clearing of the cache with `-C` to get answers that you've already searched for to appear with the `-a` flag instead of saying `Sorry, couldn't find any help with that topic`. I think we might need to version the FileSystemCache when we find breaking changes like this...

gleitz:
Actually, I stand corrected. We already [version the cache key](https://github.com/gleitz/howdoi/blob/master/howdoi/howdoi.py#L453-L454) with the semantic version of the application so when I bump it we should be good to go!

PR #316, opened 2020-10-05
Stylistic Issue

atharvakale343:
* Add comment to the line "if __name__ == '__main__':" in howdoi.py
* Add doc string to command_line_runner() function in howdoi.py
* Reverses the order of arguments in _format_answers() from res,args to args,res to maintain uniformity

P.S: I'm a student in high school. This is my first pull request. Excuse my mistakes if any.


gleitz:
Thanks for this PR, Atharva. Your suggestion to switch the `args` variable to be first is pretty good. Sometimes I do something like `code, args` and other times I do `args, links` and I think that having consistency is a good idea. Because in Python we often use `self` as the first argument I think it makes sense to pass `args` there instead.

I think that while your additional comments are useful for those beginning to write programs, they aren't necessary in the codebase.

I would like to accept this PR if you would make these changes:
1. Remove the extra comments
2. Change `def _format_output(code, args):` to `def _format_output(args, code):`
3. Change `def _format_answers(res, args):` to `def _format_answers(args, res):` (this is done I believe)

When you're done, be sure to run flake8 and pylint over the code to make sure there are no stylistic errors (for example there should be a space between `args` and `res` like this `(args, res)`.

Let me know what you think, and if you need any help along the way.

atharvakale343:
Hi Benjamin, @gleitz 
Thanks for taking the time to write an insightful note about my PR. I have made the suggested changes.

When I ran Flake8, I found this:
Flake8:
E402 module level import not at top of file (in several lines)
E501 line too long (99 > 79 characters) (in several lines)

I am not sure if these are significant stylistic issues.
Pylint ran fine.

Thanks again,
Atharva.

gleitz:
Nice work! This looks good. I'll make some changes so that flake8 runs properly (it looks like the config file should be named `.flake8` rather than `.flake8rc`. Will also set up github actions to run on PRs so we can get a nice green checkmark here and know that merges are good to go.

Thanks again for the help.

PR #318, opened 2020-10-19
Add architecture ppc64le to travis build

ddeka2910:
Thank you for the code.
Add support for architecture ppc64le.  
This is part of the Ubuntu distribution for ppc64le. This helps us simplify testing later when distributions are re-building and re-releasing. For more info tag @gerrith3

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,17 @@ matrix:
     - python: "3.8"
       dist: xenial    # required for Python 3.7 (travis-ci/travis-ci#9069)
       sudo: required  # required for Python 3.7 (travis-ci/travis-ci#9069)
+    #powerjobs
+    - python: "3.5"
+      arch: ppc64le
+    - python: "3.6"
+      arch: ppc64le
+    - python: "3.7"
+      arch: ppc64le
+    - python: "3.8"
+      arch: ppc64le
+      dist: xenial    # required for Python 3.7 (travis-ci/travis-ci#9069)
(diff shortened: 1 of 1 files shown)


gleitz:
Thanks for the PR. Is there reason to think that the tests would fail on `ppc64le`? Would you like to package this for Ubuntu?
cc @gerrith3

gerrith3:
So, obviously your question is based on the fact that python is typically architecture independent.  Which, it is, right up until it isn't.  ;)  So, any failures on another platform are typically caused by references to other python modules or binaries which can be referenced in the code.  Regression testing on other architectures at least notifies you when any such contribution happens.  In terms of packaging, I'm fine with letting canonical do their packaging/match their dependencies.  The goal is just to reduce the possibly failures at the time each distro is build by ensuring that the package runs on (at least) ppc64le.  BTW, this package is new to me, but looks cool.  I don't know if I consider myself a hack, but I have forgotten more details of various commands than I care to admit over the years, and this looks like a helpful tool for that!  ;)

gleitz:
Thanks. I had not considered that issue before. 

Are there other architectures that might be worth a test? 

If there was a test failure, should I let the offending package maintainer know, or is that something you like to see? 

Enjoy howdoi!

gerrith3:
Many are also adding arm and s390x support both of which are available at Travis today.

gerrith3:
On test failures - you can start with the offending package maintainer who will probably have more context on their project, but feel free to tag me as well.

gleitz:
Closing this because I had to switch away from Travis to Gitlab runners.

Issue #293, opened 2020-06-29
Add multiple requirement files


gmelodie:
For now we only use one `requirements.txt` file. As the project is growing (plugins, better linting and testing, new docs) I think we could split the requirements file into multiple files. Here's a structure I'm thinking about (stole from [this comment](https://stackoverflow.com/a/20720019/6032037)):

```
-- howdoi_root
|-- requirements
|   |-- common.txt
|   |-- dev.txt
|   |-- prod.txt
-- requirements.txt
```

hmanalai:
I have created a PR for this issue that adds the requirements folder and the stated .txt files. As of now, common.txt contains all the requirements in requirements.txt.

PR #327, opened 2021-01-15
Requirement Files 

hmanalai:
- This PR fixes issue #293 by adding a requirements folder that contains three different .txt files for common/development/production specific requirements.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,10 +1,2 @@
 # when adding a new dependency, also add to setup.py's `install_requires`
-Pygments>=2.3.1
-argparse==1.4.0
-cssselect==1.1.0
-lxml>=4.2.5
-pyquery==1.4.1
-requests==2.24.0
-cachelib==0.1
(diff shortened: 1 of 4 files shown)


gleitz:
Thanks for this. What do you think about linking `pip install -r requirements.txt` to `requirements/prod.txt` so we don't have any duplication?

hmanalai:
I believe this change would need to happen in the python.yml file?

gleitz:
I'm not sure about that yml file. Do you have a link to that usage? 

I think by default most people do `pip install -r requirements.txt` so I wanted to avoid having to update requirements in multiple places.

gleitz:
So my ask would be to edit the requirements.txt file to point to your new prod.txt file.

hmanalai:
I updated requirements.txt and tested it to make sure it works. Now it runs `-r requirements/prod.txt`.

gleitz:
Thanks!

Issue #338, opened 2021-02-20
Searching of query > howdoi print in python does`t work but using "howdoi print in pythonn" works 


HussainMadarwala53:
![howdoi](https://user-images.githubusercontent.com/70691828/108597218-e85ede80-73ad-11eb-9da8-7cc04d90eb7d.PNG)

gleitz:
Would you let me know what version you're using with `-v` and also try clearing the cache with `-C` and letting me know if that fixes the query? 

We've been having a bit of trouble with Google so you may also want to try `-e bing` to see if those results are better until we can fix the Google issue in the next release.

gleitz:
This should be fixed in version 2.0.13

HussainMadarwala53:
Great, Thank you so much.

Issue #332, opened 2021-02-18
Couldn't find any help with that topic on every query


SaurusXI:
Persistent issue on every query which means that functionality is completely impaired at the moment. 
Did a little digging by making the following modification to source - 
```python
def get_proxies():
    proxies = getproxies()
    print(f'Proxies {proxies}')
    ...
```
Which gives the following output - 
```
Proxies {}
Sorry, couldn't find any help with that topic
```

Looks like no proxies are being fetched by the `getproxies()` call which is creating the issue.

gleitz:
I believe this was an issue with Google search results which should now be fixed in 2.0.12. Would you please update howdoi using pip and let me know if you now get results?

dsh1705:
updated to 2.0.12 and still have the issue noted above.

PR #340, opened 2021-02-23
switch to regex for extracting links for google query

### LINKED ISSUE #339: Sorry, couldn't find any help with that topic?
every QUERY results in the same: Sorry, couldn't find any help with that topic?

anything wrong?

SaurusXI:
Switch to regex to avoid breakage every time XPATH of query results is slightly altered. Should fix #339 #338 #332 etc
Fetches links in order and works on all the examples I've tried.

### DIFF
--- a/howdoi/howdoi.py
+++ b/howdoi/howdoi.py
@@ -228,12 +228,10 @@ def _clean_google_link(link):
     return link
 
 
-def _extract_links_from_google(html):
-    links = [a.attrib['href'] for a in html('.yuRUbf')('a')] or \
-        [a.attrib['href'] for a in html('.kCrYT')('a')] or \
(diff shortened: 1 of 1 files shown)


gleitz:
It's interesting to note that the build failed due to a pylint error I had not seen before.

With a little digging I found:
https://github.com/PyCQA/pylint/issues/4133

This seems to suggest that we can remove the `[]` from the `any` query to get better performance (negligible here, but we might as well do it).

I'll fix that in a separate PR as this change LGTM.

gleitz:
Build fixed in https://github.com/gleitz/howdoi/actions/runs/593312306

gleitz:
Nice work on your first PR!! 🎉

gleitz:
Released as `2.0.13`

PR #341, opened 2021-02-23
fix: make google query extract only stackoverflow links from page

SaurusXI:
Make link extraction regex for Google extract only stackoverflow links from page, as opposed to all links


gleitz:
Thanks for this. I'm seeing a linting error so I wonder if we might need to format the regex a little differently:

```W605 invalid escape sequence '\.'```

SaurusXI:
@gleitz so not using raw strings for stuff like regex is deprecated behaviour - my bad. should be fixed now though, please take a look.

gleitz:
Cool!

Issue #337, opened 2021-02-19
Sanity check if howdoi still works and try to automatically fix it via bot


atodorov:
Related to the recent troubles with Google changing their class names and the code not being able to find any results:

- [ ] Add a sanity check that will verify that howdoi still works, something like `howdoi <...>` & assert on the result will work
- [ ] Integrate said check into CI. IIRC GitHub Actions can be triggered periodically via cron. This will serve as an early warning system. If it fails maintainers and everyone subscribed on the repository will receive a notification.
- [ ] Follow up with some code which tries to find the correct class names, could be a script initially. 
- [ ] Instrument said script to open pull requests via the GitHub API, some examples on http://atodorov.org/blog/2015/12/01/commit-a-file-with-the-github-api-and-python/, the PyGithub library probably already contains better interface to this
- [ ] If we think we trust the steps above (or if we see that they are OK the majority of the time) then implement them as an auto-fix bot of sorts: sanity test fails => runs the fixer script => opens a PR => if PR passes we could have some automation to merge the changes. 
- [ ] All of this can be hooked into GitHub Actions as a custom action instead of having to host it somewhere.

gleitz:
@SaurusXI we can probably start with a cron (once a day?) for all three engines (bing, duckduckgo, and google) with some sort of alerting (email?) that lets us know of a failure.

PR #345, opened 2021-02-26
add sanity check

SaurusXI:
Adds a `--sanity-check` flag to howdoi that checks if a standard query ("format date bash") works or fails on all currently supported engines. 
Implements the first feature in the checklist at #337

### DIFF
--- a/howdoi/howdoi.py
+++ b/howdoi/howdoi.py
@@ -39,6 +39,7 @@ from requests.exceptions import SSLError
 from howdoi import __version__
 from howdoi.errors import GoogleValidationError, BingValidationError, DDGValidationError
 
+
 # rudimentary standardized 3-level log output
 def _print_err(err):
     print("[ERROR] " + err)
@@ -640,7 +641,7 @@ def get_parser():
     return parser
 
 
-def sanity_check(test_query=None):
+def _sanity_check(test_query=None):
     parser = get_parser()
     if not test_query:
         test_query = 'format date bash'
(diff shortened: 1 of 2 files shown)


howdoi/howdoi.py
```diff
@@ -635,9 +636,50 @@ def get_parser():
...
+def _sanity_check(test_query=None):
+    parser = get_parser()
+    if not test_query:
+        test_query = 'format date bash'
+    error_result = b"Sorry, couldn't find any help with that topic\n"
+
+    if _clear_cache():
+        _print_ok('Cache cleared successfully')
+    else:
+        _print_err('Clearing cache failed')
```
atodorov, line 653: I think we should always clear the cache here.

howdoi/howdoi.py
```diff
@@ -677,6 +734,10 @@ def command_line_runner():  # pylint: disable=too-many-return-statements,too-man
         _print_ok(__version__)
         return
 
+    if args['sanity_check']:
+        perform_sanity_check()
+        return
```
atodorov, line 739: Since we want this to be executed in CI we need to inform CI of failures somehow. The easiest way is to return non-zero in case of failure. See if a `sys.exit()` call with the return value of `perform_sanity_check()` will work here.

howdoi/howdoi.py
atodorov: Resolved.

howdoi/howdoi.py
SaurusXI: This should be sorted in the latest commit I pushed, please check it out

atodorov (approved):
lgtm

PR #349, opened 2021-02-26
Include extra packages to allow MkDocs to build properly

gleitz:


.github/workflows/docs.yml
```diff
@@ -1,10 +1,10 @@
 # This is a basic workflow to help you get started with Actions
 
-name: CI
+name: MkDocs CI
 
 on:
   push:
-    branches: [ master ]
+    branches: [ master, docs ]
```
gleitz, line 7: @gmelodie do you think all branches should build the docs, or just master?

PR #350, opened 2021-03-01
add early warning system CI

### LINKED ISSUE #337: Sanity check if howdoi still works and try to automatically fix it via bot

SaurusXI:
Implements an early warning CI as discussed in #337 
The remaining features discussed in the issue are based on the previous scraping approach and might not be relevant now, so this might close #337

### DIFF
--- a/.github/workflows/early-warning.yml
+++ b/.github/workflows/early-warning.yml
@@ -1,0 +1,42 @@
+name: Early Warning Notifier
+
+on:
+  push:
+    branches: master
+  release:
+    types: [published]
+  schedule:
+    - cron: "0 0 * * *"
+
+jobs:
+  sanity-check:
+    runs-on: ubuntu-latest
+    name: Sanity Check
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+
+      - name: Setup environment
+        run: pip install -r requirements.txt
+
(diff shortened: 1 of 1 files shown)


gleitz:
Looking good @SaurusXI 
https://github.com/gleitz/howdoi/runs/2044807444?check_suite_focus=true

Issue #351, opened 2021-03-01
howdoi doesnt work


sonuishaq67:
![Screenshot from 2021-03-01 19-59-16](https://user-images.githubusercontent.com/55189255/109511012-a7ab4780-7ac8-11eb-8edb-fbe9d28449e0.png)

error log
https://pastebin.ubuntu.com/p/gtKB28Br3N/

sonuishaq67:
OS arch linux 
python version : Python 3.9.2

gleitz:
@Srinivasa314 was having the same issue, so consolidating the bug reports here.

gleitz:
@sonuishaq67 can you confirm if this error happens for every query, or just some queries?

Could you also try clearing your cache with the `-C` flag?

While we're debugging this issue, I think you should be able to use bing or duckduckgo to fetch results by changing the search engine with `-e bing`.

cc @SaurusXI

sonuishaq67:
> -e bing , duckduckgo

it works for bing and duckduckgo 

> can you confirm if this error happens for every query, or just some queries?

happens for every query with google

> -C flag

I cleared the cache but its the same for google

sonuishaq67:
@gleitz  It's working now when I clear the cache

sonuishaq67:
It works sometimes when I clear the cache and sometimes it doesn't

gleitz:
If you can share other queries and error responses that would be useful. We're trying to track down that the issue with the URL might be.

sonuishaq67:
I hope this helps 
https://pastebin.ubuntu.com/p/5FSnZQDyVH/

gleitz:
We have a fix checked into the master branch that I believe should fix this issue. Would you mind checking out howdoi and testing it for us?

```
git clone git@github.com:gleitz/howdoi.git
cd howdoi
pip3 install -r requirements.txt
```

Then try a few queries using this format

```
python -m howdoi.howdoi print in c -C
```

And then let me know if you run into any issues. Really appreciate the help!

sonuishaq67:
Yep this works perfectly :) 
Thanks for fixing.

Will be closing this.

gleitz:
Fixed in 2.0.14

PR #353, opened 2021-03-04
fix: make regex pattern more specific to avoid incorrect link matches

SaurusXI:
refer - https://meta.stackexchange.com/questions/108949/question-url-slugs-based-on-title


SaurusXI:
Should resolve occasional breakage as seen in #351

gleitz:
Added https://github.com/gleitz/howdoi/commit/22c50f18f1feb96b13c12d5ad56ec7bd086fb8b6 just to be sure there are no duplicates

Issue #354, opened 2021-03-05
Parallelization 


gleitz:
When you ask for multiple answers with `-n5` we should request those in parallel

Q: How do we preserve the order?

This doesn't work but might be close:
```    
from multiprocessing import Pool

def _get_answer_worker(answer_number):
        current_position = answer_number + initial_position
        args['pos'] = current_position
        link = get_link_at_pos(question_links, current_position)
        answer = _get_answer(args, question_links)
        if not answer:
            return ''
        if not args['link'] and not args['json_output'] and multiple_answers:
            answer = ANSWER_HEADER.format(link, answer, STAR_HEADER)
        answer += '\n'
        return answer

    with Pool() as pool:
        result = pool.map(_get_answer_worker, range(args['num_answers']))
        print(result)
```

gleitz:
cc @SaurusXI

SaurusXI:
@gleitz lets close this issue?

gleitz:
Fixed in #357

PR #357, opened 2021-03-09
Parallelize page fetch queries

SaurusXI:
Parallelizes page queries according to #354

### DIFF
--- a/howdoi/howdoi.py
+++ b/howdoi/howdoi.py
@@ -20,6 +20,9 @@ import textwrap
 from urllib.request import getproxies
 from urllib.parse import quote as url_quote, urlparse, parse_qs
 
+from multiprocessing import Pool
+from functools import partial
+
 import appdirs
 import requests
 
@@ -332,11 +335,7 @@ def _get_questions(links):
     return [link for link in links if _is_question(link)]
 
 
-def _get_answer(args, links):
-    link = get_link_at_pos(links, args['pos'])
-    if not link:
-        return False
-
+def _get_answer(link, args):
     cache_key = link
     page = cache.get(link)  # pylint: disable=assignment-from-none
     if not page:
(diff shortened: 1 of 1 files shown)


howdoi/howdoi.py
```diff
@@ -332,11 +335,7 @@ def _get_questions(links):
     return [link for link in links if _is_question(link)]
 
 
-def _get_answer(args, links):
-    link = get_link_at_pos(links, args['pos'])
-    if not link:
-        return False
-
+def _get_answer(link, args):
```
gleitz, line 338: A lot of the `howdoi` functions (`_parse_cmd`, `_format_output`, etc) take `args` as the first argument so probably best to keep the same here.

howdoi/howdoi.py
```diff
@@ -410,29 +409,41 @@ def _get_answers(args):
     if not question_links:
         return False
 
-    answers = []
-    initial_position = args['pos']
-    multiple_answers = (args['num_answers'] > 1 or args['all'])
+    initial_position = args['pos'] - 1
+    final_position = initial_position + args['num_answers']
+    question_links = question_links[initial_position: final_position]
```
gleitz, line 414: I think you can remove the whitespace between `initial_position: final_position`

howdoi/howdoi.py
```diff
@@ -410,29 +409,41 @@ def _get_answers(args):
...
-        if not args['link'] and not args['json_output'] and multiple_answers:
-            answer = ANSWER_HEADER.format(link, answer, STAR_HEADER)
-        answer += '\n'
-        answers.append({
-            'answer': answer,
-            'link': link,
-            'position': current_position
-        })
+    for idx, _ in enumerate(answers):
+        answers[idx]['position'] = idx + 1
```
gleitz, line 420: This is clever, I was wondering how you were doing to keep them in order.

howdoi/howdoi.py
```diff
@@ -410,29 +409,41 @@ def _get_answers(args):
...
 
     return answers
 
 
+def _get_answer_worker(link, args):
```
gleitz, line 425: I suppose this can also take `args` as the first parameter.

gleitz (approved):
Just a few nits and then this looks good to merge.

howdoi/howdoi.py
SaurusXI: Yeah the threadpool maps elements one-one so ordering was preserved, but indexing from inside a parallel section is a recipe for disaster - so I had to resort to assigning position labels after the fact :smile:

howdoi/howdoi.py
SaurusXI: Right, will push a commit for these changes in a bit

howdoi/howdoi.py
SaurusXI: I think I'll also change the variable names, line 414 seems a little too wordy

SaurusXI:
Push changes according to the nits. Please let me know if any more changes are needed

PR #358, opened 2021-03-26
Cache patch

### LINKED ISSUE #355: Cache is not cleared on new release
All calls to the cache should use `_get_cache_key` so that `__version__` is appended.

`_get_answer` and `_get_links_with_cache` do not currently use it.

SaurusXI:
Closes #355 

Just FYI - adding inspect calls seem to slow cached queries down by 10ms on my local machine

### DIFF
--- a/howdoi/howdoi.py
+++ b/howdoi/howdoi.py
@@ -335,7 +335,7 @@ def _get_questions(links):
 
 
 def _get_answer(args, link):
-    cache_key = link
+    cache_key = _get_cache_key(link)
     page = cache.get(link)  # pylint: disable=assignment-from-none
     if not page:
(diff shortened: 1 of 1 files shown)


howdoi/howdoi.py
```diff
@@ -376,7 +377,7 @@ def _get_answer(args, link):
 
 
 def _get_links_with_cache(query):
-    cache_key = query + "-links"
+    cache_key = _get_cache_key(query + "-links")
```
gleitz, line 380: Here we can remove the `-links` because it was only there to do what `calling_func` does now.

SaurusXI: Ah right, I'll make that change

Issue #334, opened 2021-02-19
Standardized Logging


gleitz:
- Standardize the way that errors are logged and reported to the user
- Get rid of `_print_err` and use something standard
- Verbosity level can then be controlled by command line flags
- `Explainability` can perhaps be sent to standard error? Is this a common practice or should it go to standard out.

gmelodie:
Related to #330

atodorov:
@gleitz do you mind assigning this to MLH fellow @KimaruThagna

gleitz:
Yes, please! Thanks @KimaruThagna

KimaruThagna:
Taken it up. Will revert with ideas and work done

gleitz:
Proposed PR: https://github.com/gleitz/howdoi/pull/381

PR #359, opened 2021-03-30
Standardized logging

KimaruThagna:
Links to issue #334

### DIFF
--- a/howdoi/errors.py
+++ b/howdoi/errors.py
@@ -1,5 +1,6 @@
 import logging
 
+
 class GoogleValidationError(Exception):
 
     logging.exception(f'Google Validation Error{Exception}')
--- a/howdoi/errors.py
+++ b/howdoi/errors.py
@@ -6,5 +6,6 @@ class GoogleValidationError(Exception):
 class BingValidationError(Exception):
     pass
 
+
 class DDGValidationError(Exception):
     pass
--- a/howdoi/errors.py
+++ b/howdoi/errors.py
@@ -1,18 +1,10 @@
-import logging
 
 
 class GoogleValidationError(Exception):
-
-    def __init__(self):
-        logging.exception('Google Validation Error %s ', Exception)
-
+    pass
 
 class BingValidationError(Exception):
-
-    def __init__(self):
-        logging.exception('Bing Validation Error %s ', Exception)
-
+    pass
 
 class DDGValidationError(Exception):
-    def __init__(self):
-        logging.exception('DDG Validation Error %s', Exception)
+    pass
--- a/howdoi/errors.py
+++ b/howdoi/errors.py
@@ -3,6 +3,7 @@
 class GoogleValidationError(Exception):
     pass
 
+
 class BingValidationError(Exception):
     pass
 
--- a/howdoi/errors.py
+++ b/howdoi/errors.py
@@ -3,12 +3,12 @@ import logging
 
 class GoogleValidationError(Exception):
 
-    logging.exception(f'Google Validation Error{Exception}')  # pylint: disable=Use lazy % formatting in logging functions (logging-fstring-interpolation)
+    logging.exception('Google Validation Error %s ', Exception)
 
 
 class BingValidationError(Exception):
-    logging.exception(f'Bing Validation Error{Exception}')  # pylint: disable=Use lazy % formatting in logging functions (logging-fstring-interpolation)
+    logging.exception('Bing Validation Error %s ', Exception)
 
 
 class DDGValidationError(Exception):
-    logging.exception(f'DDG Validation Error{Exception}')  #  pylint: disable=Use lazy % formatting in logging functions (logging-fstring-interpolation)
+    logging.exception('DDG Validation Error %s', Exception)
--- a/howdoi/errors.py
+++ b/howdoi/errors.py
@@ -3,12 +3,12 @@ import logging
 
 class GoogleValidationError(Exception):
 
-    logging.exception(f'Google Validation Error{Exception}')
+    logging.exception(f'Google Validation Error{Exception}')  # pylint: disable=Use lazy % formatting in logging functions (logging-fstring-interpolation)
 
 
 class BingValidationError(Exception):
-    logging.exception(f'Bing Validation Error{Exception}')
+    logging.exception(f'Bing Validation Error{Exception}')  # pylint: disable=Use lazy % formatting in logging functions (logging-fstring-interpolation)
 
 
 class DDGValidationError(Exception):
-    logging.exception(f'DDG Validation Error{Exception}')
+    logging.exception(f'DDG Validation Error{Exception}')  #  pylint: disable=Use lazy % formatting in logging functions (logging-fstring-interpolation)
--- a/howdoi/errors.py
+++ b/howdoi/errors.py
@@ -3,12 +3,16 @@ import logging
 
 class GoogleValidationError(Exception):
 
-    logging.exception('Google Validation Error %s ', Exception)
+    def __init__(self):
+        logging.exception('Google Validation Error %s ', Exception)
 
 
 class BingValidationError(Exception):
-    logging.exception('Bing Validation Error %s ', Exception)
+
+    def __init__(self):
+        logging.exception('Bing Validation Error %s ', Exception)
 
 
 class DDGValidationError(Exception):
-    logging.exception('DDG Validation Error %s', Exception)
+    def __init__(self):
+        logging.exception('DDG Validation Error %s', Exception)
--- a/howdoi/errors.py
+++ b/howdoi/errors.py
@@ -1,10 +1,13 @@
+import logging
+
 class GoogleValidationError(Exception):
-    pass
+
+    logging.exception(f'Google Validation Error{Exception}')
 
 
 class BingValidationError(Exception):
-    pass
+    logging.exception(f'Bing Validation Error{Exception}')
 
 
 class DDGValidationError(Exception):
-    pass
+    logging.exception(f'DDG Validation Error{Exception}')
--- a/howdoi/howdoi.py
+++ b/howdoi/howdoi.py
@@ -423,7 +423,7 @@ def _get_answers(args):
     search_engine = os.getenv('HOWDOI_SEARCH_ENGINE', 'google')
 
     info_logging('{} links found on {}: {}'.format(URL, search_engine, len(question_links)))
-    logging.info('Answers requested: {} starting at position: {}'.format(args['num_answers'], initial_pos))
+    info_logging('Answers requested: {} starting at position: {}'.format(args['num_answers'], initial_pos))
 
     with Pool() as pool:
         answers = pool.starmap(
--- a/howdoi/howdoi.py
+++ b/howdoi/howdoi.py
@@ -23,6 +23,7 @@ from urllib.parse import quote as url_quote, urlparse, parse_qs
 
 from multiprocessing import Pool
 
+import logging
 import appdirs
 import requests
 
@@ -41,7 +42,7 @@ from requests.exceptions import SSLError
 
 from howdoi import __version__
 from howdoi.errors import GoogleValidationError, BingValidationError, DDGValidationError
-import logging
+
 
 
 def error_logging(err):  # log level=error
--- a/howdoi/howdoi.py
+++ b/howdoi/howdoi.py
@@ -328,11 +328,11 @@ def _get_answer(args, link):  # pylint: disable=too-many-branches
     cache_key = link
     page = cache.get(link)  # pylint: disable=assignment-from-none
     if not page:
-        logging.warning('Fetching page: %s', link)
+        print(f'Fetching page: {link}')
         page = _get_result(link + '?answertab=votes')
         cache.set(cache_key, page)
     else:
-        logging.warning('Using cached page: %s', link)
+        print(f'Using cached page: {link}')
 
     html = pq(page)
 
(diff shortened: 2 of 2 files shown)


howdoi/howdoi.py
```diff
@@ -41,17 +43,19 @@
 from howdoi import __version__
 from howdoi.errors import GoogleValidationError, BingValidationError, DDGValidationError
 
+logging.basicConfig(format='%(levelname)s: %(message)s')
+
 
-# rudimentary standardized 3-level log output
-def _print_err(err):
-    print(f'[ERROR] {err}')
+def error_logging(err):  # log level=error
```
gleitz, line 49: I think we can remove these three functions in favor of a simple call to `logging.error` or whatever the underlying function will be.

howdoi/errors.py
```diff
@@ -1,10 +1,14 @@
+import logging
+
+
 class GoogleValidationError(Exception):
-    pass
+
+    logging.exception('Google Validation Error %s ', Exception)
```
gleitz, line 6: I believe this code will run when the file is compiled? I see a lot of this in the logs:

```
ERROR:root:Bing Validation Error <class 'Exception'>
```

Instead, we will only want to print this out when the class is used.

howdoi/howdoi.py
```diff
@@ -512,7 +526,7 @@ def print_stash(stash_list=None):
         stash_list = ['\nSTASH LIST:']
         commands = keep_utils.read_commands()
         if commands is None or len(commands.items()) == 0:
-            print(f'No commands found in stash. Add a command with "howdoi --{STASH_SAVE} <query>".')
+            debug_logging(f'No commands found in stash. Add a command with "howdoi --{STASH_SAVE} <query>".')
```
gleitz, line 529: I believe we'll want to use a different log level here. Previously, the print statement would always be shown when trying to view and empty stash. With this PR it isn't shown anymore when you try:

```
python -m howdoi.howdoi --empty
python -m howdoi.howdoi --view
```

It could be an error log?

howdoi/howdoi.py
```diff
@@ -534,9 +548,9 @@ def _stash_remove(cmd_key, title):
     commands = keep_utils.read_commands()
     if commands is not None and cmd_key in commands:
         keep_utils.remove_command(cmd_key)
-        print(f'\n{BOLD}{GREEN}"{title}" removed from stash.{END_FORMAT}\n')
+        info_logging(f'\n{BOLD}{GREEN}"{title}" removed from stash.{END_FORMAT}\n')
     else:
-        print(f'\n{BOLD}{RED}"{title}" not found in stash.{END_FORMAT}\n')
+        debug_logging(f'\n{BOLD}{RED}"{title}" not found in stash.{END_FORMAT}\n')
```
gleitz, line 553: This might also want to be an error log?

howdoi/howdoi.py
```diff
@@ -578,12 +592,15 @@ def howdoi(raw_query):
...
+        info_logging('Using cached response (add -C to clear the cache)')
         return _parse_cmd(args, res)
 
+    info_logging('Fetching answers for query: {}'.format(args['query']))
+
     try:
         res = _get_answers(args)
         if not res:
-            res = {'error': 'Sorry, couldn\'t find any help with that topic\n'}
+            res = {'error': 'Sorry, couldn\'t find any help with that topic\n(use --explain to learn why)'}
```
gleitz, line 603: We should check if the `--explain` flag is set before including this message.

howdoi/howdoi.py
```diff
@@ -640,9 +658,9 @@ def _sanity_check(test_query=None):
     error_result = b"Sorry, couldn't find any help with that topic\n"
 
     if _clear_cache():
-        _print_ok('Cache cleared successfully')
+        info_logging('Cache cleared successfully')
```
gleitz, line 661: Again, this message is not shown. Let's figure out what log level to use.

howdoi/howdoi.py
```diff
@@ -706,15 +724,15 @@ def perform_sanity_check():
...
     except BingValidationError:
-        _print_err('Bing query failed')
+        error_logging('Bing query failed')
         return -1
     except DDGValidationError:
-        _print_err('DuckDuckGo query failed')
+        error_logging('DuckDuckGo query failed')
         return -1
-    print('Ok')
+    info_logging('Sanity check passed')
```
gleitz, line 735: This might need to be a different log level

howdoi/howdoi.py
```diff
@@ -723,19 +741,23 @@ def command_line_runner():  # pylint: disable=too-many-return-statements,too-man
...
+
     if args['sanity_check']:
         sys.exit(
             perform_sanity_check()
         )
 
     if args['clear_cache']:
         if _clear_cache():
-            _print_ok('Cache cleared successfully')
+            info_logging('Cache cleared successfully')
```
gleitz, line 758: Needs a different log level

gleitz (commented):
Getting closer! Added a few comments

howdoi/errors.py
gleitz: Call this in the`__init__`

howdoi/howdoi.py
```diff
@@ -534,9 +548,9 @@ def _stash_remove(cmd_key, title):
     commands = keep_utils.read_commands()
     if commands is not None and cmd_key in commands:
         keep_utils.remove_command(cmd_key)
-        print(f'\n{BOLD}{GREEN}"{title}" removed from stash.{END_FORMAT}\n')
+        info_logging(f'\n{BOLD}{GREEN}"{title}" removed from stash.{END_FORMAT}\n')
```
gleitz, line 551: warning

howdoi/howdoi.py
gleitz: warning

howdoi/errors.py
KimaruThagna: Changing this to use constructors

howdoi/howdoi.py
KimaruThagna: Done

howdoi/howdoi.py
KimaruThagna: will be using logical AND. So this message will only be set if not res AND -x is false

gleitz:
We're getting closer. At this point, I would have you try a few queries and see what the output is like:

```
 » python -m howdoi.howdoi format date bash
WARNING: Fetching answers for query: format date bash
WARNING: stackoverflow.com links found on google: 1
WARNING: Answers requested: 1 starting at position: 0
WARNING: Fetching page: https://stackoverflow.com/questions/1401482/yyyy-mm-dd-format-date-in-shell-script
WARNING: Total answers returned: 1
# put current date as yyyy-mm-dd in $date
# -1 -> explicit current date, bash >=4.3 defaults to current time if not provided
# -2 -> start time for shell
printf -v date '%(%Y-%m-%d)T\n' -1

# put current date as yyyy-mm-dd HH:MM:SS in $date
printf -v date '%(%Y-%m-%d %H:%M:%S)T\n' -1

# to print directly remove -v flag, as such:
printf '%(%Y-%m-%d)T\n' -1
# -> current date printed to terminal
```

This shows me that the logging level for these warning messages is not correct. I think we can break it down like this:

- error - an error
- warning - general output. I believe this is replacing most `print` statements. @atodorov I could use your advice here on whether we should keep print statements in the code or make them all warnings.
- info - all of the output for explainability (many of the logging statements in this PR are warnings and I believe they should be info instead so they are not shown unless you pass `--explain`)

gleitz:
@KimaruThagna would you mind going thru each output statement, think about the level chosen, and if you have an questions you can @mention me in the comments? 

I think that because this is a command line application, we can use the `print` statement for the main program output. Everything for explainability can be info. Errors are error. And warning we can look at on a case-by-case basis. 

Thx!

KimaruThagna:
@gleitz in the previous PR remember we changed the info log level to warning. 
I believe we should go back to print statements 
We leave logging for actual events such as error. 
You previously has reservations about the info log level. Where you requested me to change all of them  to warning. Has this changed? I also agree that info should be for explainability. This leaves the warning log level. I think it should be for mild errors.

KimaruThagna:
@gleitz Checkout the new changes. All program outputs are prints except for errors. Errors are being handled by` logging.error` The explainability part is handled by `info` as you've setup the log level in the code. I have not seen any tangible use case for warning. For the print statements, any that involves a success callback of sorts I have given them a green font

howdoi/howdoi.py
```diff
@@ -335,12 +325,15 @@ def _get_questions(links):
     return [link for link in links if _is_question(link)]
 
 
-def _get_answer(args, link):
-    cache_key = _get_cache_key(link)
+def _get_answer(args, link):  # pylint: disable=too-many-branches
+    cache_key = link
     page = cache.get(link)  # pylint: disable=assignment-from-none
     if not page:
+        print(f'Fetching page: {link}')
```
gleitz, line 332: We only want to display this when explainability is enabled.

howdoi/howdoi.py
```diff
@@ -335,12 +325,15 @@ def _get_questions(links):
...
         page = _get_result(link + '?answertab=votes')
         cache.set(cache_key, page)
+    else:
+        print(f'Using cached page: {link}')
```
gleitz, line 336: We only want to display this when explainability is enabled.

howdoi/howdoi.py
```diff
@@ -357,8 +350,10 @@ def _get_answer(args, link):
         answer_body_cls = ".post-text"
 
     if not instructions and not args['all']:
+        print('No code sample found, returning entire answer')
```
gleitz, line 353: We only want to display this when explainability is enabled.

howdoi/howdoi.py
```diff
@@ -357,8 +350,10 @@ def _get_answer(args, link):
...
         text = get_text(first_answer.find(answer_body_cls).eq(0))
     elif args['all']:
+        print('Returning entire answer')
```
gleitz, line 356: We only want to display this when explainability is enabled.

howdoi/howdoi.py
```diff
@@ -371,6 +366,7 @@ def _get_answer(args, link):
     else:
         text = _format_output(args, get_text(instructions.eq(0)))
     if text is None:
+        print(f'{RED}Answer was empty')
```
gleitz, line 369: We only want to display this when explainability is enabled.

howdoi/howdoi.py
```diff
@@ -412,6 +408,10 @@ def _get_answers(args):
     initial_pos = args['pos'] - 1
     final_pos = initial_pos + args['num_answers']
     question_links = question_links[initial_pos:final_pos]
+    search_engine = os.getenv('HOWDOI_SEARCH_ENGINE', 'google')
+
+    print(f'{URL} links found on {search_engine}: {len(question_links)}')
+    print(f'Answers requested: {args["num_answers"]} starting at position: {initial_pos}')
```
gleitz, line 414: We only want to display these when explainability is enabled.

howdoi/howdoi.py
```diff
@@ -422,6 +422,7 @@ def _get_answers(args):
     for idx, _ in enumerate(answers):
         answers[idx]['position'] = idx + 1
 
+    print(f'Total answers returned: {len(answers)}')
```
gleitz, line 425: We only want to display this when explainability is enabled.

howdoi/howdoi.py
```diff
@@ -581,12 +582,15 @@ def howdoi(raw_query):
...
+        logging.info('Using cached response (add -C to clear the cache)')
         return _parse_cmd(args, res)
 
+    print(f'Fetching answers for query: {args["query"]}')
+
     try:
         res = _get_answers(args)
-        if not res:
-            res = {'error': 'Sorry, couldn\'t find any help with that topic\n'}
+        if not res and not args["explain"]:  # only set message when they havent set the -x flag
```
gleitz, line 592: Here, we always want to return `Sorry, couldn't find any help with that topic`

However, if explain was not used, we will want to append to this string `(use --explain to learn why)`

howdoi/howdoi.py
```diff
@@ -726,19 +731,23 @@ def command_line_runner():  # pylint: disable=too-many-return-statements,too-man
     args = vars(parser.parse_args())
 
     if args['version']:
-        _print_ok(__version__)
+        print(__version__)
         return
 
+    if args['explain']:
+        logging.getLogger().setLevel(logging.INFO)
+        print('Version: %s', __version__)
```
gleitz, line 739: Let's go ahead and use logging.info here as well for the version.

howdoi/howdoi.py
gleitz: Remember: explainability statements will be logged with `logging.info` because we don't want this information to be printed unless the `--explain` flag is set.

howdoi/howdoi.py
```diff
@@ -537,9 +538,9 @@ def _stash_remove(cmd_key, title):
     commands = keep_utils.read_commands()
     if commands is not None and cmd_key in commands:
         keep_utils.remove_command(cmd_key)
-        print(f'\n{BOLD}{GREEN}"{title}" removed from stash.{END_FORMAT}\n')
+        print(f'\n{BOLD}{GREEN} {title} removed from stash.{END_FORMAT}\n')
```
gleitz, line 541: Any reason to remove the `"` here? I think it's useful to show the name of query that was removed (or not found) in the stash.

howdoi/howdoi.py
```diff
@@ -643,9 +648,9 @@ def _sanity_check(test_query=None):
     error_result = b"Sorry, couldn't find any help with that topic\n"
 
     if _clear_cache():
-        _print_ok('Cache cleared successfully')
+        print(f'{GREEN}Cache cleared successfully')
```
gleitz, line 651: When using `GREEN` you must also use `END_FORMAT`.

For example right now this entire query is green:

```
python -m howdoi.howdoi format date bash -C
```

KimaruThagna:
Now things are getting abit more clear. Before none of this was brought to
the front but now I have a clearer picture of what is expected. Will work
on it.

gleitz:
@KimaruThagna thanks for the help. Would like to get the last issues fixed so we can help the folks who are having trouble in https://github.com/gleitz/howdoi/issues/372

gleitz:
Closed in favor of https://github.com/gleitz/howdoi/pull/381

PR #370, opened 2021-04-05
Keep checking remaining engines if some engine fails 

### LINKED ISSUE #366: option --sanity-check only checking first engine
the first engine to be checked is `google` which fails and then the others engine are not checked, would be nice if all engines are checked even if one fails.

adbenitez:
close #366

### DIFF
--- a/howdoi/howdoi.py
+++ b/howdoi/howdoi.py
@@ -694,6 +694,7 @@ def perform_sanity_check():
     exit_code = 0
     for engine in ('google', 'bing', 'duckduckgo'):
         print('Checking {}...'.format(engine))
+        os.environ['HOWDOI_SEARCH_ENGINE'] = engine
         try:
             _sanity_check(engine)
         except (GoogleValidationError, BingValidationError, DDGValidationError):
--- a/howdoi/howdoi.py
+++ b/howdoi/howdoi.py
@@ -643,13 +643,12 @@ def _sanity_check(engine, test_query=None):
     parser = get_parser()
     if not test_query:
         test_query = 'format date bash'
-    error_result = b"Sorry, couldn't find any help with that topic\n"
 
-    args = vars(parser.parse_args(test_query.split()))
+    args = vars(parser.parse_args(('-j ' + test_query).split()))
     args['search_engine'] = engine
 
     try:
-        assert howdoi(args).encode('utf-8', 'ignore') != error_result
+        assert isinstance(howdoi(args).encode('utf-8', 'ignore'), list)
     except AssertionError as exc:
         if engine == 'google':
             raise GoogleValidationError from exc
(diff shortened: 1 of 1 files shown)


adbenitez:
found two bugs in `perform_sanity_check()`:

* argument query was not split by word so the query was processed like a list of letters
* always `google` engine was tried despite `args['search_engine']` being different because the engine was set in `command_line_runner()` so subsequent calls to `howdoi()` keep using the same engine.

adbenitez:
fixed another bug that was giving wrong `Ok` result when actually the machine was completely offline

gleitz:
Awesome! Thanks this is a nice improvement.

gleitz:
Hmm, I merged this and immediately saw failures so there must be something weird with it.

https://github.com/gleitz/howdoi/runs/2292710834?check_suite_focus=true

Going to try on my local machine...

gleitz:
```
» python -m howdoi --sanity-check
Checking google...
[ERROR] google query failed
Checking bing...
[ERROR] bing query failed
Checking duckduckgo...
[ERROR] duckduckgo query failed
```

adbenitez:
sadly that probably only means the check is finally working :'D

gleitz:
So I tried printing:

```
print(howdoi(args).encode('utf-8', 'ignore'))
```

And it comes back as 

```
b'[{"answer": "# put current date as yyyy-mm-dd in $date\\n# -1 -> explicit current date, bash >=4.3 defaults to current time if not provided\\n# -2 -> start time for shell\\nprintf -v date \'%(%Y-%m-%d)T\\\\n\' -1 \\n\\n# put current date as yyyy-mm-dd HH:MM:SS in $date\\nprintf -v date \'%(%Y-%m-%d %H:%M:%S)T\\\\n\' -1 \\n\\n# to print directly remove -v flag, as such:\\nprintf \'%(%Y-%m-%d)T\\\\n\' -1\\n# -> current date printed to terminal\\n", "link": "https://stackoverflow.com/questions/1401482/yyyy-mm-dd-format-date-in-shell-script", "position": 1}]'
```

So it's not a list, but a bytes type. Very odd...

adbenitez:
ohhhhh

gleitz:
So yeah I think this was related to your earlier comment about how `b"Sorry, couldn't find any help with that topic\n"` is not a good response with we use `-j`.

gleitz:
Is this too gross?

```
        result = howdoi(args)
        assert "Sorry" not in result
```

adbenitez:
I forgot the output was bytes not json object, yikes

gleitz:
I'm thinking something like https://github.com/gleitz/howdoi/pull/375

adbenitez:
> Is this too gross?
> 
> ```
>         result = howdoi(args)
>         assert "Sorry" not in result
> ```

the issue is that there is an error message like:

```
b'Unable to reach google. Do you need to use a proxy?\n'
```
so I could just check for the two error messages, it will be still better than old code

PR #374, opened 2021-04-07
check for HTTP errors (ex. 403), to better detect when a server is blocking us

### LINKED ISSUE #367: DuckDuckGo block not detected
I executed `howdoi` like 3 times with cache disabled while testing in my laptop and DuckDuckGo is returning this instead of html:

```
If this error persists, please let us know: error-lite@duckduckgo.com
```
this is not in the `BLOCK_INDICATORS`

adbenitez:
complements BLOCK_INDICATORS.

close #367


gleitz:
Looking good!

PR #375, opened 2021-04-08
An option for fixing the sanity check issue

gleitz:


adbenitez:
see https://github.com/gleitz/howdoi/pull/370#issuecomment-815380684

still there is an error not detected

gleitz:
Good call. Added both of those error strings and a note to switch to a JSON based check when we make the `-j` flag respect errors.

gleitz:
Yeah it's more of a bandaid until we have a more robust solution

PR #376, opened 2021-04-08
Explainability

kingpitka:
Three changes, two of them are simple, small typos in the index.md and user-guide.md.
The third is changing how the error message throws "(use --explain to learn why)", previously it threw it in all cases where "Sorry, couldn't find any help with with that topic", now it only throws it if you're not already using "--explain".


gleitz:
Thanks! I especially appreciate the copy editing.

I'm going to move to https://github.com/gleitz/howdoi/pull/381 so would you mind rebasing your PR against that version?

Issue #266, opened 2020-06-10
NLP Language Detection


HajerMahjoub:
using a model that  capable of identifying entities related from the query and parsing the query and extract the important entity using POS part of the speech   and ENR entity name recognition (verb: write,install,program ...;language: c, python,c#,java ....;noun: string )

rkbeatss:
@HajerMahjoub i can help with this
![bitmoji](https://sdk.bitmoji.com/render/panel/a790712f-94fd-4c40-95ab-fec033bd90e5-abe04bbd-ecc6-4ea5-b453-bbccd5c9d443-v1.png?transparent=1&palette=1&width=246)

HajerMahjoub:
@rkbeatss will be great working with you again

gleitz:
I'd start here with a high level story about what can be improved with ML.

It may we that we want to tackle refinement. For example, if someone searches for `howdoi print stack trace` we get the java result back. Could we suggest other terms like `python` or `javascript` that might be attached to a query? What else could we do to improve the stackoverflow experience?

If may be that we want to tackle choosing the correct underlying provider of the answer. For example a recipe provider vs wikihow vs stackoverflow. Given that 99% of people are probably expecting the stackoverflow answer, how can we continue to provide those users a great experience who also allowing for other sites to be included?

HajerMahjoub:
I will work on that goal starting from today
I have couple of ideas using skills similarity model to suggest similar topic @gleitz  as the example you set I will use rasa to extract entities to better understand the resource we need to fulfill the answer we need  if it's Wiki or stackoverflow or recipe  website

gleitz:
@SaurusXI

gleitz:
Problem: print stack trace returns results for PHP but you often ask questions about Python
Goal: howdoi (a) recognizes that the query is missing a language and (b) suggests additional answers based on your top used languages

Pieces:
- binary classifier (print stack trace -> isLanguage vs. configure web server -> isNotLanguage)
- regex language extractor (print stack trace php -> php)
- personal stats and language preference (store a tally of the tags)

gleitz:
Another avenue of exploration is knowing when to return a code block or not. @SaurusXI suggested that if we had a non-language query, you probably want the whole answer.

We could also look at the ratio of the length of the code block vs. the entire answer to know if we should return it or not.

For example `howdoi parse html in a regex` should return the entire answer.

PR #380, opened 2021-04-13
feat: add language labelling notebook so far

SaurusXI:
Dump of the NLP notebook so far.


SaurusXI:
addresses the NLP binary classifier discussed in #266

PR #392, opened 2021-05-30
Added Lint command and pre-commit hook

Prateek93a:
I have added both `lint` command and pre-commit hook, in case someboady wants to run `python setup.py lint` manually without adding the modified files to staging area.

Also for the command, currently I have added basic messages for the case when no linting errors are found and for the case when `flake8` is not installed. Any suggestions for improving these messages would be very helpful.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,4 +11,4 @@ repos:
     hooks:
       - id: pylint
         args:
-          - '--rcfile=.pylintrc'
+          - "--rcfile=.pylintrc"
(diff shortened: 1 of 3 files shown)


gleitz:
This is great. I was wondering how to do this. My only suggestion would be to run pylint as well, as that is the other check that runs when you open a PR and it would be good to know if either fails. I think you can just extend the `Lint` command rather than adding a new one.

Prateek93a:
Thanks for the feedback! I will make the additions and update the PR soon.

Prateek93a:
Currently this is how the text appears when we use the `lint` command ->
- For no errors
ezgif com-gif-maker (1)

- For errors
ezgif com-gif-maker (2)


So, is it okay or does it need more formatting?

gleitz:
Looking good! Thanks again

Issue #407, opened 2021-07-12
Search engine failure fallback


ShezaMunir:
Add in functionality that takes care of temporary block failures. 
In case google fails, howdoi should revert to bing or duckduckgo and vice versa for all three search engines.

gleitz:
This was fixed in #408

PR #408, opened 2021-07-19
Search engine fallbacks - including test 

ShezaMunir:
Addresses issue #407 . 

Fallbacks in case of search engine failure: iterates through the supported engines to find a search engine that has not been blocked. If all supported engines are blocked, it simply returns. 

image

### DIFF
--- a/howdoi/howdoi.py
+++ b/howdoi/howdoi.py
@@ -591,8 +591,8 @@ def howdoi(raw_query):
     else:
         args = raw_query
 
-    os.environ['HOWDOI_SEARCH_ENGINE'] = args['search_engine'] or os.getenv('HOWDOI_SEARCH_ENGINE') or 'google'
-    search_engine = os.getenv('HOWDOI_SEARCH_ENGINE')
+    search_engine = args['search_engine'] or os.getenv('HOWDOI_SEARCH_ENGINE') or 'google'
+    os.environ['HOWDOI_SEARCH_ENGINE'] = search_engine
     if search_engine not in SUPPORTED_SEARCH_ENGINES:
         supported_search_engines = ', '.join(SUPPORTED_SEARCH_ENGINES)
         message = f'Unsupported engine {search_engine}. The supported engines are: {supported_search_engines}'
@@ -622,9 +622,9 @@ def howdoi(raw_query):
             res = {'error': message}
         cache.set(cache_key, res)
     except (RequestsConnectionError, SSLError):
-        res = {'error': f'Unable to reach {args["search_engine"]}. Do you need to use a proxy?\n'}
+        res = {'error': f'Unable to reach {search_engine}. Do you need to use a proxy?\n'}
     except BlockError:
-        BLOCKED_ENGINES.append(args['search_engine'])
+        BLOCKED_ENGINES.append(search_engine)
         next_engine = next((engine for engine in SUPPORTED_SEARCH_ENGINES if engine not in BLOCKED_ENGINES), None)
         if next_engine is None:
             res = {'error': 'Unable to get a response from any search engine\n'}
--- a/howdoi/howdoi.py
+++ b/howdoi/howdoi.py
@@ -626,7 +626,7 @@ def howdoi(raw_query):
     except BlockError:
         BLOCKED_ENGINES.append(args['search_engine'])
         if BLOCKED_ENGINES == SUPPORTED_SEARCH_ENGINES:
-            return
+            return howdoi(args)
         for eng in SUPPORTED_SEARCH_ENGINES:
             if eng not in BLOCKED_ENGINES:
                 logging.error('%sRetrying search with %s%s', GREEN, eng, END_FORMAT)
--- a/howdoi/howdoi.py
+++ b/howdoi/howdoi.py
@@ -585,11 +585,9 @@ def _parse_cmd(args, res):
 
 
 def howdoi(raw_query):
-    if isinstance(raw_query, str):  # you can pass either a raw or a parsed query
-        parser = get_parser()
-        args = vars(parser.parse_args(raw_query.split(' ')))
-    else:
-        args = raw_query
+    if isinstance(raw_query['query'], str):  # you can pass either a raw or a parsed query
+        raw_query['query'] = raw_query['query'].split(' ')
+    args = raw_query
 
     os.environ['HOWDOI_SEARCH_ENGINE'] = args['search_engine'] or os.getenv('HOWDOI_SEARCH_ENGINE') or 'google'
     search_engine = os.getenv('HOWDOI_SEARCH_ENGINE')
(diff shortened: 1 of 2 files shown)


howdoi/howdoi.py
```diff
@@ -621,7 +623,16 @@ def howdoi(raw_query):
         cache.set(cache_key, res)
     except (RequestsConnectionError, SSLError):
         res = {'error': f'Unable to reach {args["search_engine"]}. Do you need to use a proxy?\n'}
-
+    except BlockError:
+        BLOCKED_ENGINES.append(args['search_engine'])
+        if BLOCKED_ENGINES == SUPPORTED_SEARCH_ENGINES:
```
gleitz, line 628: I don't believe that this equality check will work between `BLOCKED_ENGINES`, a list and `SUPPORTED_SEARCH_ENGINES`, a tuple.

```
>>> type((1, 2, 3))
<class 'tuple'>
>>> type([1, 2, 3])
<class 'list'>
>>> (1, 2, 3) == [1, 2, 3]
False
```

However, if you made both of these objects a `Set` then the equality will work

```
>>> set((1, 2, 3)) == set([1, 2, 3])
True
```

You can probably change to

```
SUPPORTED_SEARCH_ENGINES = set(['google', 'bing', 'duckduckgo'])
BLOCKED_ENGINES = set()
```

rather than doing the conversion only when you do the comparison.

howdoi/howdoi.py
```diff
@@ -621,7 +623,16 @@ def howdoi(raw_query):
...
+            return howdoi(args)
```
gleitz, line 629: If these sets are now equal, which means that we've exhausted all the engines, when we want to return an error statement, right? The code as written today might cause an infinite loop.

howdoi/howdoi.py
```diff
@@ -621,7 +623,16 @@ def howdoi(raw_query):
...
+        for eng in SUPPORTED_SEARCH_ENGINES:
+            if eng not in BLOCKED_ENGINES:
+                logging.error('%sRetrying search with %s%s', GREEN, eng, END_FORMAT)
```
gleitz, line 632: Let's use `logging.info` here since it's not quite an error yet since we have a fallback we can use.

howdoi/howdoi.py
```diff
@@ -621,7 +623,16 @@ def howdoi(raw_query):
         cache.set(cache_key, res)
     except (RequestsConnectionError, SSLError):
         res = {'error': f'Unable to reach {args["search_engine"]}. Do you need to use a proxy?\n'}
-
+    except BlockError:
+        BLOCKED_ENGINES.append(args['search_engine'])
+        if BLOCKED_ENGINES == SUPPORTED_SEARCH_ENGINES:
+            return howdoi(args)
+        for eng in SUPPORTED_SEARCH_ENGINES:
```
gleitz, line 630: Another way to write this could be 

```
[engine for engine in SUPPORTED_SEARCH_ENGINES if engine not in BLOCKED_ENGINES]
```

However, you then must reverse the list because if you started with ('google', 'bing', 'duckduckgo') you're going to get back ('duckduckgo', 'bing', 'google') and we probably want to try these search engines in the order they are written.

What do you think about:

```
next_search_engine = next(reversed([engine for engine in SUPPORTED_SEARCH_ENGINES if engine not in BLOCKED_ENGINES]), None)
```

You can check the result of this this, if it's `None` then you can print the error. If it's not then you can use that as the next search engine to try.

howdoi/howdoi.py
gleitz: Note - look at my comment after this - we might be able to remove the equality check altogether.

gleitz:
I wonder why we're getting 

```
INFO: Fetching answers for query: f o r m a t   d a t e   b a s h
```

On the retry?

howdoi/howdoi.py
ShezaMunir: Yes, I tried doing just a `return` statement here, instead of `return howdoi(args)`, but that was throwing a pylint error (inconsistent return statements) because the rest of the return statements are returning stuff. Will look into this

howdoi/howdoi.py
ShezaMunir: You're right, it's not going to work with sets and lists. I've updated the method so that it doesn't use the if statement here, using the next() function instead as suggested below.

howdoi/howdoi.py
ShezaMunir: Yep, I thought that too, but added in the `return howdoi(args)` as simply `return` was throwing a pylint error. I've taken the alternate route with the `next()` function now. However, I'd like to test the code to see what happens when all search engines fail and whether it is the expected output (updated now). Any ideas on how to go about that (breaking stuff on purpose)?

howdoi/howdoi.py
ShezaMunir: So I tried using `logging.info` too, but `INFO` logs only show up when you use `--explain` which normally the user doesn't go for. If you don't use the `--explain` argument, you see that the search engine blocked your request, but you still got the answer (the rerouting to Google isn't visible as `logging.info` will not show up).  I've left it as `logging.error` but let me know if `logging.info` would still be the preferred route, I'll change it to that

howdoi/howdoi.py
ShezaMunir: Thanks for sharing this! I was looking for a more compact method to iterate through the list and find unblocked engines, one at a time, and this is perfect for that. I've added it in now, let me know what you think

ShezaMunir:
> I wonder why we're getting
> 
> ```
> INFO: Fetching answers for query: f o r m a t   d a t e   b a s h
> ```
> 
> On the retry?
> 
> It looks like something is putting spaces in between the letters in the search query as well, so this is more than just a cosmetic issue.

#### The fix I tried
Before checking the web for stack overflow links, howdoi joins the `raw_query['query']` list, adding spaces between each word. But the problem here was that when the query was a string instead of a list, the code to split it into a list wasn't being accessed. The `isIntance` function's argument was a dict (`raw_query`) instead of the query (`raw_query['query']`) which would be a string or a list. My solution is working, but it is not passing the nose test (throwing TypeError), can you please suggest a solution?

#### Previous code
The previous code, which was using the parser, even when accessed wasn't solving the problem. The implementation would split the string into a list but it kept changing the search engine to `None` instead of the fallbacks one.

ShezaMunir:
Ready to be merged :) @gleitz

ShezaMunir:
The test is ready for review @gleitz 

A couple of things I'd like to discuss:
1. When I ran the lint checks on my PC, they came out fine, but for some reason, the GitHub checks threw errors. Is that usual? What might be the reason for that?
2. When I change the name of the test function, it doesn't work properly and throws errors. I've googled but couldn't find a reason for that, I'd like to learn more about this.

PR #411, opened 2021-07-22
fix a bug

hongkuancn:


gleitz:
That does indeed look like a bug. Would you mind explaining how you found it and steps to reproduce so I can verify?

hongkuancn:
I was learning the source code and happened to find the bug. To reproduce it, I think you can use the same query with different arguments, and check the log info.

For example, to run the following commands
```
python -m howdoi -C print hello world in python
python -m howdoi -j print hello world in python
```

Because the two commands have the same query, we should get the same stackoverflow link, and the second command can use the cached page generated by the first command. 

After modifying the code, we can see the following log entry once instead of twice when running the above commands.
```
INFO: Fetching page: https://stackoverflow.com/questions/1077347/hello-world-in-python
````

gleitz:
Thanks! Excellent find.

PR #423, opened 2021-09-09
Show error message when we aren't able to find an answer

gleitz:
Fixes #383

PR #424, opened 2021-09-11
filter empty results from answers

justin-f-perez:
* fix "total answers" logging message counting empty results
* return False from `_get_answers` if the only results had no answer


justin-f-perez:
@gleitz i think this fixes the logging problem mentioned in #423, and i also noticed the docstring specifies 'return False if no answers'

gleitz:
Thanks!

PR #427, opened 2021-09-23
Drop python2 support in command_line_runner

### LINKED ISSUE #426: remove Python2 support in command_line_runner
### Description 
README.md says,
> Unfortunately Python 2.7 support has been discontinued :(

I think we can drop support for the older version of Python [here](https://github.com/gleitz/howdoi/blob/f19ca79c7aff69e09c5393b2ede6ab24cddf6f75/howdoi/howdoi.py#L816).

I am not sure if we still need utf-8 encoding.
Maybe we can do like this?

```
utf8_result = howdoi(args).encode('utf-8', 'ignore')
sys.stdout.buffer.write(utf8_result)

# or
result = howdoi(args)
print(result)
```

### Have you read the Contributing Guidelines on Pull Requests on mkdocs?

Yes

ykskks:
This PR is related to #426 

- removed if statement for checking python version
- removed utf-8 encoding

close #426

### DIFF
--- a/howdoi/howdoi.py
+++ b/howdoi/howdoi.py
@@ -812,12 +812,9 @@ def command_line_runner():  # pylint: disable=too-many-return-statements,too-man
     if os.getenv('HOWDOI_COLORIZE'):
         args['color'] = True
 
-    utf8_result = howdoi(args).encode('utf-8', 'ignore')
-    if sys.version < '3':
-        print(utf8_result)
-    else:
-        # Write UTF-8 to stdout: https://stackoverflow.com/a/3603160
-        sys.stdout.buffer.write(utf8_result)
+    result = howdoi(args)
+    print(result)
+
     # close the session to release connection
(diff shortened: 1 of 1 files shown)


gleitz:
I think we want to take the `else` from the `if` statement, because that is the path that supports `> 3`.

```
utf8_result = howdoi(args).encode('utf-8', 'ignore')
# Write UTF-8 to stdout: https://stackoverflow.com/a/3603160
sys.stdout.buffer.write(utf8_result)
```

ykskks:
I wasn't sure why we needed to encode to utf-8, but now I probably understand.

If the stdout encoding is somehow not utf-8 in user's environment, then some non-ascii characters might not be printed. Thus we force the utf-8 encoding in stdout.

I will update the code.

gleitz:
Yes, I believe that's correct. We arrived at this through some trial and error over time.

Issue #431, opened 2021-10-04
Improved answer layout in the terminal


gleitz:
### Description 
Somewhat similar to https://github.com/gleitz/howdoi/issues/430, could we have a better layout in the terminal when you're picking multiple answers, or looking at different positional answers. 

Might get some inspiration from https://github.com/santinic/how2

gleitz:
This is fixed with the `rich` additions https://github.com/gleitz/howdoi/commit/88873b5c5ee000986fd71242ba069b05547e7e5e

PR #443, opened 2021-11-23
Syntax highlighting using rich library

V2dha:
## Description:

- Tasks solved

  - Added syntax highlighting feature using rich library ~~(still needed to hardcode language)~~

- Links to issues solved - [431](https://github.com/gleitz/howdoi/issues/431)

## Pull Request type:

- New feature
- Improvement

## Pull Request checklist:

- [x] Read the [contributing_to_howdoi.md](https://github.com/gleitz/howdoi/blob/master/docs/contributing_to_howdoi.md)
- [x] Attach screenshots of expected behavior.
- [x] The changes pass tests locally (`nosetests`).
- [x] There are no linting errors (`python setup.py lint`).
- [x] The changes don't break existing features.
- [x] Check that there are no confidential files like `.env` included.
- [x] Request review from the maintainers.
- [x] For bug fixes or changes to directory structure, make sure docs are updated.

## Known bugs (if any):

The function currently has hardcoded "python" language as its language parameter. I tried using lexer as a global variable in the _format_output function to add as the language parameter but it returned None.

Capture

### DIFF
--- a/howdoi/howdoi.py
+++ b/howdoi/howdoi.py
@@ -316,12 +316,10 @@ def get_link_at_pos(links, position):
         link = links[-1]
     return link
 
-lexer = None
 
 def _format_output(args, code):
-    if not args['color']:    
+    if not args['color']:
         return code
-    global lexer
 
     # try to find a lexer using the StackOverflow tags
     # or the query arguments
@@ -342,6 +340,7 @@ def _format_output(args, code):
     return highlight(code,
                      lexer,
                      TerminalFormatter(bg='dark'))
+
 
 def _is_question(link):
     for fragment in BLOCKED_QUESTION_FRAGMENTS:
@@ -817,11 +816,7 @@ def command_line_runner():  # pylint: disable=too-many-return-statements,too-man
     result = howdoi(args)
     # Write UTF-8 to stdout: https://stackoverflow.com/a/3603160
     # sys.stdout.buffer.write(utf8_result)
-    syntax =  Syntax(result,
-                  lexer,   
-                  background_color = "default",  
-                  line_numbers=False)
-    
+    syntax = Syntax(result, "python", background_color="default", line_numbers=False)
     console = Console()
     console.print(syntax)
 
--- a/howdoi/howdoi.py
+++ b/howdoi/howdoi.py
@@ -814,13 +814,12 @@ def command_line_runner():  # pylint: disable=too-many-return-statements,too-man
         args['color'] = True
 
     result = howdoi(args)
-    # Write UTF-8 to stdout: https://stackoverflow.com/a/3603160
-    # sys.stdout.buffer.write(utf8_result)
-    syntax = Syntax(result, "python", background_color="default", line_numbers=False)
+    lang = guess_lexer(result).name
+    syntax = Syntax(result, lang, background_color="default", line_numbers=False)
     console = Console()
     console.print(syntax)
 
-    # close the session to release connection
+    # close the session to release connectione
     howdoi_session.close()
 
 
--- a/howdoi/howdoi.py
+++ b/howdoi/howdoi.py
@@ -819,7 +819,7 @@ def command_line_runner():  # pylint: disable=too-many-return-statements,too-man
     console = Console()
     console.print(syntax)
 
-    # close the session to release connectione
+    # close the session to release connection
     howdoi_session.close()
 
 
--- a/howdoi/howdoi.py
+++ b/howdoi/howdoi.py
@@ -337,8 +337,9 @@ def _format_output(args, code):
 
     syntax = Syntax(code, lexer, background_color="default", line_numbers=False)
     console = Console(record=True)
-    console.print(syntax)
-    return console.export_text(styles=True)
+    with console.capture() as capture:
+        console.print(syntax)
+    return capture.get()
 
 
 def _is_question(link):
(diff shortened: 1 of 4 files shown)


V2dha:
The main problem right now that I am facing is how to detect the language of the returned answer in the syntax highlighting.
I tried to use the pygment's guess_lexer function so I set the lexer to global variable to see the lexer was returned upon passing the args["tags"] and args["query"] but it returned None.
Another things I tried was to extract the language out of args["query"] and set it as the language for the code.

V2dha:
I have updated the PR. Now instead of hardcoding the default language as "Python". I am using Pygment's guess_lexer function to detect the source code language of the answers returned.
Below are some examples -

Capture
Capture
Capture

gleitz:
Glad to see this proof of concept working!

Rather than using the syntax highlighting on the entire answer, what do you think about instead replacing only the [`highlight` function](https://github.com/gleitz/howdoi/blob/master/howdoi/howdoi.py#L339)? You can then use the proper `lexer` as well.

This will also fix the issue where `howdoi format date bash -n3` tries to highlight the entire answer rather than just the code snippets.

Screen Shot 2021-11-30 at 8 19 04 PM

Also, you'll need to add the `rich` library to the [requirements file](https://github.com/gleitz/howdoi/blob//requirements/common.txt#L2) (replacing Pygments) and also the [`setup.py`](https://github.com/gleitz/howdoi/blob//setup.py#L99).

V2dha:
Yes, earlier I replaced pygment's `highlight` function with rich's syntax function but in the end it wasn't fully leveraging the syntax highlighting feature as shown below.
Capture


I don't know the full reason but maybe it's due to the fact that the syntax object keeps reinitiating every time the function `_format_output` is called and it prints it like that on the rich's console.

Also, right now this query `howdoi format date bash -n3` gives the following output 

Capture

I have added rich library in the requirements.txt and setup.py but I haven't removed pygment as its `guess_lexer` function is still needed in the code.

V2dha:
I have replaced the `syntax` function with `highlight` function.

However, it fails the test `test_colorize` 
Capture

gleitz:
Ah interesting. That test is looking for the control character `[39;` which is probably not used by rich.

I wonder if there is a way to get the raw control characters that are printed to the terminal so we can see what kind of colors rich uses in that case.

V2dha:
I was thinking that since now it is using rich for syntax highlighting can we remove the `test_colorize` test or is it associated with some other functionalities.

gleitz:
I do think we want to keep that test because we should only see colorization when you pass that flag. If you don't it shouldn't colorize anything. This is used mostly by people who work in terminals that don't support colorization (and who don't want to see a bunch of `[39;` in their code)

V2dha:
But then with syntax highlighting it would always highlight the code irrespecitve of the flag `-c`

gleitz:
If you don't set the `-c` flag, then `_format_output` will just return without calling `highlight`. The same can happen for the `syntax` function.

https://github.com/gleitz/howdoi/blob/master/howdoi/howdoi.py#L319-L320

gleitz:
@V2dha Check out this section on [Capturing output](https://rich.readthedocs.io/en/stable/console.html#capturing-output).

I created this patch that seems to do the right thing – allows us to capture the output in `_format_output` without printing it to the console, and then prints everything at the end.

```
diff --git a/howdoi/howdoi.py b/howdoi/howdoi.py
index 978717fb..61e763e1 100644
--- a/howdoi/howdoi.py
+++ b/howdoi/howdoi.py
@@ -337,8 +337,9 @@ def _format_output(args, code):
 
     syntax = Syntax(code, lexer, background_color="default", line_numbers=False)
     console = Console(record=True)
-    console.print(syntax)
-    return console.export_text(styles=True)
+    with console.capture() as capture:
+        console.print(syntax)
+    return capture.get()
 
 
 def _is_question(link):
@@ -813,11 +814,7 @@ def command_line_runner():  # pylint: disable=too-many-return-statements,too-man
         args['color'] = True
 
     result = howdoi(args)
-    if args['color']:
-        console = Console()
-        console.print(result)
-    else:
-        sys.stdout.buffer.write(result.encode('utf-8', 'ignore'))
+    sys.stdout.buffer.write(result.encode('utf-8', 'ignore'))
 
     # close the session to release connection
     howdoi_session.close()
```

You can apply this by saving the file and then running `git apply console-patch-from-gleitz.patch`

`rich` does seem to highlight a bit better (on the right) than pygments (on the left).

Screen Shot 2021-12-12 at 9 43 21 PM

V2dha:
I have applied the `console-patch-from-gleitz.patch` but the nosetests seems to fail with `test_colorize` whereas when I run the `test_howdoi.py` on my system it passes.
Capture

gleitz:
Alright sorry for the delay. Here's how I think we might go about debugging this.

First, we need to see what control characters are actually being printed. In the failing test you can add:

```
        print(f'\nNormal:\n{repr(normal)}')
        print(f'\nColorized:\n{repr(colorized)}')
```

Then you can run that test with `nosetests -v --nocapture test_howdoi:HowdoiTestCase.test_colorize`.

For me, it passes.

Next, we'll want to see why this test fails when the CI is run. It could be that `rich` uses different colorization depending on the type of console that is being printed to. Perhaps you can I are using a dark theme, and the CI is using a light theme.

So you'll want to modify https://github.com/gleitz/howdoi/blob/master/.github/workflows/python-non-master.yml#L39 to instead run `nosetests -v --nocapture`, push that to your branch, and then check the output. We should then be able to see what control characters are being used within the CI.

Hopefully there is some common overlap between a control character used within the CI and our local machines!

gleitz:
Fixed the last issue with tests in https://github.com/gleitz/howdoi/pull/453

See [this commit](https://github.com/gleitz/howdoi/pull/453/commits/c53b6a179a09159740de2c06fb87b194e810f839) for an explanation of the fix specific for Github Actions.

